Issue #7, opened 2019-10-17
Strongly Typed SDKs


bign8:
## Describe the proposal
Hey folks!  Saw this stuff was released recently and am super excited about it!

One thing I noticed while looking through these SDKs is that they seem to just be exposing the generated gRPC code of the desired language.  I was wondering if it's worth wrapping the gRPC logic in some thin, language specific wrappers.  This could allow language specific patterns to be exposed in each of the SDKs rather than expecting all consumers to understand gRPC right out of the gate.

I threw a branch together of something that it _could_ look like, but wanted to get some feedback on the idea from the maintainers before I go down a rabbit hole.

https://github.com/dapr/go-sdk/compare/master...bign8:language-specific-api

Looking forward to collaborating!

(Also, sorry if this comes across and edgy or short, I have yet to have my morning coffee ☕️)

amanbha:
@bign8 that's great.
Generated sdks are just the beginning, we plan to have strongly typed sdks. Looking forward to collaboration.

amanbha:
@bign8 Perfect, we can start with your PR.
cc @yaron2 as well.

mchmarny:
PTAL at the "bespoke client proposal" PR #18

mchmarny:
Closing this since #18 landed in master

PR #8, opened 2019-10-17
[wip] Strongly Typed Go SDK

bign8:
# Description

> One thing I noticed while looking through these SDKs is that they seem to just be exposing the generated gRPC code of the desired language. I was wondering if it's worth wrapping the gRPC logic in some thin, language specific wrappers. This could allow language specific patterns to be exposed in each of the SDKs rather than expecting all consumers to understand gRPC right out of the gate.
> \- https://github.com/dapr/go-sdk/issues/7

_TODO: better PR description (NOTE: PR is in draft state)_

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: https://github.com/dapr/go-sdk/issues/7

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Specification has been updated
* [ ] Provided sample for the feature

### DIFF
--- a/example/caller/main.go
+++ b/example/caller/main.go
@@ -3,46 +3,34 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
 
-	pb "github.com/dapr/go-sdk/dapr"
-	"github.com/golang/protobuf/ptypes/any"
-	"google.golang.org/grpc"
+	"github.com/dapr/go-sdk/pkg/dapr"
+	"github.com/golang/protobuf/ptypes/wrappers"
 )
 
 func main() {
-	// Get the Dapr port and create a connection
-	daprPort := os.Getenv("DAPR_GRPC_PORT")
-	daprAddress := fmt.Sprintf("localhost:%s", daprPort)
-	conn, err := grpc.Dial(daprAddress, grpc.WithInsecure())
+
+	// Create the client
+	client, err := dapr.NewClient()
 	if err != nil {
 		fmt.Println(err)
 		return
 	}
-	defer conn.Close()
-
-	// Create the client
-	client := pb.NewDaprClient(conn)
+	defer client.Close()
 
 	// Invoke a method called MyMethod on another Dapr enabled service with id client
-	resp, err := client.InvokeService(context.Background(), &pb.InvokeServiceEnvelope{
-		Id:     "client",
-		Data:   &any.Any{Value: []byte("Hello")},
-		Method: "MyMethod",
-	})
+	req := &wrappers.StringValue{Value: `Hello`}
+	resp := &wrappers.StringValue{}
+	err = client.Invoke(context.Background(), `client`, `MyMethod`, req, resp)
 	if err != nil {
 		fmt.Println(err)
 	} else {
-		fmt.Println(string(resp.Data.Value))
+		fmt.Println(resp.Value)
 	}
 
 	// Publish a message to the topic TopicA
-	_, err = client.PublishEvent(context.Background(), &pb.PublishEventEnvelope{
-		Topic: "TopicA",
-		Data: &any.Any{
-			Value: []byte("Hi from Pub Sub"),
-		},
-	})
+	message := &wrappers.StringValue{Value: `Hi from Pub Sub`}
+	err = client.Publish(context.Background(), `TopicA`, message)
 	if err != nil {
 		fmt.Println(err)
 	} else {
@@ -50,15 +38,9 @@ func main() {
 	}
 
 	// Save state with the key myKey
-	_, err = client.SaveState(context.Background(), &pb.SaveStateEnvelope{
-		Requests: []*pb.StateRequest{
-			&pb.StateRequest{
-				Key: "myKey",
-				Value: &any.Any{
-					Value: []byte("My State"),
-				},
-			},
-		},
+	err = client.SaveState(context.Background(), &dapr.State{
+		Key:   `myKey`,
+		Value: &wrappers.StringValue{Value: `My State`},
 	})
 	if err != nil {
 		fmt.Println(err)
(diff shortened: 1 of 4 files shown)


bign8:
Okay folks, so far I've basically pulled as much logic I can that has to do with the gRPC definitions out of the examples and placed it in the helper package `pkg/dapr`.  I'm still getting my test environment all setup for Dapr, so no guarantees that this actually works.  But it's enough to get the conversation started on the types we are looking to expose to consumers and how the consumers interact with those types.

**Interesting decision points so far**

I'm Trying to have consumers not have to deal with `*any.Any`.  This involves using `proto.Message` and `ptypes.[Un]MarshalAny` along with needing to accept the expected model type in some cases. Not sure if this is absolutely necessary, but it may feel more natural for the general consumer.

Since only one gRPC server can be running in `client` scope, all bindings, topics and invocation handlers are bound to a global map, not great go pattern, but limited by a system restriction.  So it "might" be forgivable (Rob Pike, Russ Cox and friends... I'm sorry).

And I'm still skimming the docs for the purposes for all the values attached to various payloads.  This current abstraction hides a lot of the extra information available directly from the protobuffs. Since the exact use case of the extra metadata is not shown in the examples, I don't really know how a consumer would use them, so have chosen to omit it for now (don't mind finding ways to work them back in tho).

The `Serve` API is likely to change pending https://github.com/dapr/dapr/issues/102 or having the Dapr sidecar injector add an environment variable to the source container `DAPR_PORT`.  But currently it's accepting a hard-coded port, which if https://github.com/dapr/dapr/issues/102 merges after this does, would make the cleanup of `Serve` a "breaking" API change.  Not a huge deal while Dapr itself is in 0.1.0 phase, but should be considered before making a 1.0.0 tag.

Anyway, lots of assumptions going in here. Most issues are likely due to my ignorance around the particulars of the project.  Hopefully I'm not stepping on toes, just trying to get something out there to start discussion.

yaron2:
hey @bign8, first off, thanks for this: we actually strive to have our SDKs be hand-crafted and your PR coming in so early is great: thanks from the entire Dapr team!

I agree with your assessment regarding tests, and https://github.com/dapr/dapr/issues/102 is likely to get resolved in the next week or so.

Re: metadata, we need to still pass them, as each metadata object can carry implementation-specific data. For example, when invoking a Kafka binding, you could pass `topic:topic2` to override the topic in the Kafka component yaml.

We are working on documentation of all metadata properties for each of the service calls.

bign8:
Hey @yaron2, always happy to contribute!  This stuff is really cool and it's a privilege to be able to help with its development.  Thank you!

I took a stab at doing some optional API things for metadata and grpc call options borrowing some designs from [Dave Cheney](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis). If you link some of the documented use cases for the metadata parameters, I can make things more strongly typed for those use cases.  For the client calls, they seem semi-reasonable, but the server invocations get a little harry.

A few questions came up while I was doing this that may be worth asking (hopefully most of these are simply answered by quoting various docs).

1. It looks like there are a few options on the `BindingsResponseEnvelope` that imply that any other calls to `client.<modifier>State` should be all applied at once.  If thats the case, some sort of construct will have to be created here such that `client.<modifier>State` calls don't actually call out to the client (sidecar + dapr system) but instead buffer them until the service responds.
2. Should all calls to a dapr gRPC client be initiated from another call coming into the system. 
 Maybe this is showing my history with the [App Engine SDK](https://github.com/golang/appengine) (hopefully it's not forbidden to discuss other SDKs and such here). But appengine had an `appengine.Main` function that would start everything, and give `appengine.Contexts` out to things, that then must be used to make other calls into the system.  I'm wondering if it would be worth adopting a similar system here where `dapr.Main` could start a `dapr.DaprClient` and then methods like `Invoke`, `Publish`, `Binding`, and `<modifier>State` could get the `dapr.DaprClient` off of the context passed in.  For server/background tasks they also offered an `appengine.Background` to make a useful context for those sort of things.  Really long description just to ask if this is a pattern worth mirroring in this SDK?
3. A small extension of this is a bigger question about design: how restrictive would you folks like to see this SDK?  Given that the fallback option is just use the gRPC models directly, this is probably the spot for some more opinionated designs and call structures, but still being able to say... "well hey, you don't like it, just use the gRPC models directly".  There is always a choice in these types of things of whether to be more restrictive and flexible in the frameworks.  In the end it's really up to the core maintainers how restrictive or flexible we want to go, but I hope it ends up being a balance of the two.  Citing a goal from great philosopher of Titan:  
    
    > "Perfectly balanced, as all things should be" -[Thanos](https://youtu.be/Vx5prDjKAcw)
4. Are there documented values/units for `State.Etag`, `*.Metadata`, `StateOptions.Concurrency`, `StateOptions.Consistency`, `RetryPolicy.Pattern`, `RetryPolicy.Threshold`, etc... (basically, anything that is an int, or [Stringly typed](https://blog.codinghorror.com/new-programming-jargon/)). Hopefully I'm not beating a dead horse here, but those could be considered code smells.  Personally, I don't mind as long as they are documented somewhere, but I figured I should at least say something.  Things like `Topics` make sense here, but `Metadata` and others could maybe use enum based keys/values or strongly typed objects.

Apologies for the long form questions, just hoping to provide more of my context. Just trying to understand the design intentions fully to help make this an SDK where it's hard to shoot yourself in the foot.  Thanks again for being open to contributions and transparent discussion!

mchmarny:
@bign8 PTAL PR #18. Rather then modifying the auto-generated files, it creates a bespoke client that hides the proto types and exposes pure go API to the user. I've updated the caller example to illustrate the new usage. Let me know what you think.

bign8:
@mchmarny That looks great! not sure if the maintainers are in for vendoring dependencies or not with modules, but that looks much closer to what I was originally imagining for the client implementation.  Good way to hide the gRPC implementation away from the end user.  I'll close this in favor of #18.

*Note: I'm just a community member and not associated with the project*

PR #18, opened 2020-05-01
Custom go SDK (aka bespoke client proposal)

mchmarny:
Custom go client that addresses issue #7 and hides much of the gRPC complexity and simplifies the interfaces with which the user interact. It also updates the calling service example to use the custom client


vendor/modules.txt
```diff
@@ -0,0 +1,59 @@
+# github.com/golang/protobuf v1.3.2
```
bign8, line 1: nit: not sure if adding all vendored dependencies was indented here since `master` ([14aff55 at time of comment](https://github.com/dapr/go-sdk/tree/14aff5543faa60837ab02a1eb97cdc8a14a33455)) has a `go.mod` and no `vendor` directory.  Might suggest separating the vendoring of deps from a feature proposal.  Just an opinion, feel free to ignore.

*Note: Just a community member, not a Microsoft employee or dapr maintainer*

mchmarny: There are strong opinions on both sides of the issue re vendor dir in git, in this case, the project is voletile right now (ramping up to 0.7 release) so I vendored deps with this PR to minimize test time issues.

youngbupark:
@mchmarny There are grpc service and protobuf message definition changes in 0.8.0. Can you please update this pr accordingly? we no longer use mod vendor, either.

mchmarny:
@youngbupark the PR been tested with v0.8 and vendor dir removed

mchmarny:
BTW, I also updated the proto update script (`bin/protogen`) to eliminate the need for `protoc` and any potential issues with version diffs therein. The updated script copies the generated go files from `dapr/dapr` and replaces the import source to local path.

client/client.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+
+	pb "github.com/dapr/go-sdk/dapr/proto/runtime/v1"
+)
+
+const (
+	daprPortDefault    = "4000"
```
amanbha, line 17: change default port to 50001 (the default for kubernetes) for the NewClient() works in k8s

client/invoke.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	if method == "" {
+		return nil, errors.New("nil method")
+	}
+
+	resp, err := c.protoClient.InvokeService(authContext(ctx), &pb.InvokeServiceRequest{
+		Id: serviceID,
+		Message: &v1.InvokeRequest{
+			Method: method,
+			Data:   &any.Any{Value: in},
```
amanbha, line 26: nil check on in.

client/state.go
```diff
@@ -0,0 +1,88 @@
...
+func (c *Client) SaveStateJSON(ctx context.Context, store, key string, in interface{}) error {
+	b, err := json.Marshal(in)
+	if err != nil {
+		return errors.Wrap(err, "error marshaling content")
+	}
+	return c.SaveState(authContext(ctx), store, key, b)
+}
+
+// GetState is the message to get key-value states from specific state store
+func (c *Client) GetState(ctx context.Context, store, key string) (out []byte, err error) {
```
amanbha, line 49: need to add ConsistencyMode

amanbha: need a method which returns state value and etag as well.

client/state.go
```diff
@@ -0,0 +1,88 @@
...
+	"context"
+	"encoding/json"
+
+	v1 "github.com/dapr/go-sdk/dapr/proto/common/v1"
+	pb "github.com/dapr/go-sdk/dapr/proto/runtime/v1"
+	"github.com/pkg/errors"
+)
+
+// SaveState is the message to save multiple states into state store
+func (c *Client) SaveState(ctx context.Context, store, key string, in []byte) error {
```
amanbha, line 13: Need to add ConsistencyMode, ConcurrencyMode & RetryOptions

client/state.go
```diff
@@ -0,0 +1,88 @@
...
+	result, err := c.protoClient.GetState(authContext(ctx), envelop)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting state from %s for %s key", store, key)
+	}
+
+	return result.Data, nil
+}
+
+// DeleteState is the message to delete key-value states from specific state store
+func (c *Client) DeleteState(ctx context.Context, store, key string) error {
```
amanbha, line 70: Need to add ConsistencyMode, ConcurrencyMode & RetryOptions

client/binding.go
```diff
@@ -0,0 +1,39 @@
...
+	_, err := c.protoClient.InvokeBinding(authContext(ctx), envelop)
+	if err != nil {
+		return errors.Wrapf(err, "error invoking binding %s", name)
+	}
+
+	return nil
+}
+
+// InvokeBindingJSON invokes configured Dapr binding with an instance
+func (c *Client) InvokeBindingJSON(ctx context.Context, name, operation string, in interface{}) error {
```
amanbha, line 33: support out param for response from bidirectional bindings
it should handle 2 variations:
payload and response, payload and no response

client/invoke.go
```diff
@@ -0,0 +1,44 @@
...
+	"encoding/json"
+
+	v1 "github.com/dapr/go-sdk/dapr/proto/common/v1"
+	pb "github.com/dapr/go-sdk/dapr/proto/runtime/v1"
+	"github.com/golang/protobuf/ptypes/any"
+	"github.com/pkg/errors"
+)
+
+// InvokeService represents the request message for Service invocation
+func (c *Client) InvokeService(ctx context.Context, serviceID, method string, in []byte) (out []byte, err error) {
```
amanbha, line 14: needs to cover 4 sceanrios, please ensure all of these can be achieved here:
No Data and no return value.
No Data but a return value.
Data but no return value.
both Data and return value.

example/serving/components/pubsub.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
```
amanbha, line 2: default file created by, its not needed for running samples locally

example/serving/components/statestore.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: dapr.io/v1alpha1
```
amanbha, line 1: default file created by, its not needed for running samples locally

client/client.go
mchmarny: done

client/invoke.go
mchmarny: checked

client/state.go
mchmarny: provides additional methods where these value can be specified

client/state.go
mchmarny: provides additional methods where these value can be specified

client/state.go
mchmarny: provides additional methods where these value can be specified

client/binding.go
mchmarny: provides additional methods for all of these combinations

client/invoke.go
mchmarny: provides additional methods for all of these combinations

example/serving/components/statestore.yaml
mchmarny: Same as above, explicit > magic in examples

example/serving/components/pubsub.yaml
mchmarny: Given that the `example-http-binding` needs to be there for the output binding example to work it's better to be explicit here and not have these files magically show up for the user

client/binding.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if resp != nil {
+		return resp.Data, resp.Metadata, nil
+	}
+
+	return nil, nil, nil
+}
+
+// InvokeBindingJSON invokes configured Dapr binding with an instance
+func (c *Client) InvokeBindingJSON(ctx context.Context, name, operation string, in interface{}) (out []byte, outm map[string]string, err error) {
```
amanbha, line 37: Is a separate method needed here for OutputBinding as well.?

client/invoke.go
```diff
@@ -0,0 +1,95 @@
...
+			Data:        &anypb.Any{Value: data},
+			ContentType: contentTpe,
+		},
+	}
+
+	return c.InvokeServiceWithRequest(ctx, req)
+}
+
+// InvokeServiceJSON represents the request message for Service invocation with identity parameter
+func (c *Client) InvokeServiceJSON(ctx context.Context, serviceID, method string, in interface{}) (out []byte, err error) {
```
amanbha, line 76: What is the usecase for these *JSON methods for all the building blocks??

client/state.go
```diff
@@ -0,0 +1,164 @@
...
+	_, err := c.protoClient.SaveState(authContext(ctx), req)
+	if err != nil {
+		return errors.Wrap(err, "error saving state")
+	}
+
+	return nil
+}
+
+// SaveStateItem saves a single state item
+func (c *Client) SaveStateItem(ctx context.Context, store string, item *v1.StateItem) error {
```
amanbha, line 49: this is providing a mix and match of custom methods and underlying auto generated types. Provide wrapper types in the sdk layer, so that auto generated sdk is not bleeding through it.

client/state.go
```diff
@@ -0,0 +1,164 @@
...
+	result, err := c.protoClient.GetState(authContext(ctx), req)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting state")
+	}
+
+	return result.Data, nil
+}
+
+// GetState retreaves state from specific store using default consistency option
+func (c *Client) GetState(ctx context.Context, store, key string) (out []byte, err error) {
```
amanbha, line 117: need a method which returns state value and etag as well.

Readme.md
```diff
@@ -1,8 +1,6 @@
-# Dapr SDK for Go
+# dapr SDK for Go
```
amanbha, line 1: Use Dapr instead of dapr.

client/binding.go
mchmarny: this method accounts for all the options (no input, no output etc)

client/invoke.go
mchmarny: Allows the developer to pass an existing struct so they don't have to do the JSON serialization themselves

client/state.go
mchmarny: done

Readme.md
mchmarny: done

client/state.go
mchmarny: Etag exists on GetStateResponse and DeleteState. GetStateRequest doesn't expose get by Etag if that what you mean.

client/invoke.go
```diff
@@ -0,0 +1,95 @@
...
+	"encoding/json"
+
+	v1 "github.com/dapr/go-sdk/dapr/proto/common/v1"
+	pb "github.com/dapr/go-sdk/dapr/proto/runtime/v1"
+	anypb "github.com/golang/protobuf/ptypes/any"
+	"github.com/pkg/errors"
+)
+
+// InvokeServiceWithRequest invokes service with input request
+func (c *Client) InvokeServiceWithRequest(ctx context.Context, req *pb.InvokeServiceRequest) (out []byte, err error) {
```
amanbha, line 14: this doesn't need to be exported. its just a helper method for other InvokeService* methods

client/binding.go
amanbha: then along the similar lines, we dont need separate InvokeBinding & InvokeOutputBinding. One method can account for everything, thoughts?

client/state.go
amanbha: yes, there should be GetStateAndEtag method which returns data and ETag from GetStateResponse. Currently there is no method to get ETag.

client/state.go
```diff
@@ -0,0 +1,307 @@
...
+	req := &State{
+		StoreName: store,
+		States:    []*StateItem{item},
+	}
+
+	return c.SaveState(ctx, req)
+}
+
+// SaveStateWithData saves the data into store using default state options
+func (c *Client) SaveStateWithData(ctx context.Context, store, key string, data []byte) error {
```
amanbha, line 215: Can it take an ETag as well, else for Saving state with ETag user will always have to create a StateItem and then call SaveStateItem.
Can we look into renaming and reorganizing these method names. For Delete there is DeleteStateWithOptions & DeleteState. For SaveState, I would expect similar methods.

client/invoke.go
mchmarny: good catch, yes, this is only used internally

client/invoke.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "error marshaling in parameter")
+	}
+
+	req := &pb.InvokeServiceRequest{
+		Id: serviceID,
+		Message: &v1.InvokeRequest{
+			Method:      method,
+			Data:        &anypb.Any{Value: b},
+			ContentType: "application/json",
```
youngbupark, line 89: I would define const for application/json

youngbupark:
@mchmarny can you resolve the conflicts ?

client/state.go
mchmarny: We can certainly think through the method naming. IMO the diff in between Delete and Save is that Delete does have an option to delete using state options in the proto. In case of Save, there is Save multiple items, in which case each item has its own state option. I added Save one item to not force user to create an array just to invoke the method but this too has a state inside of the item. Maybe it's better to rethink the delete use-case?

client/invoke.go
mchmarny: I added it but why is JSON diff from other content types? Also, these are globally defined (spec) so not sure the constnating this in SDK doesn't make much sense.

client/state.go
mchmarny: done

mchmarny:
@amanbha @yaron2 Given the amount of edits in this PR, and the fact that this PR doesn't actually hyde the auto generated protos that are already available in master, is there perhaps a case for focusing on cleaning up the core methods and then do small incremental improvements to the custom client after?

client/binding.go
mchmarny: Yes, the one method can account for all scenarios. The only reason I left `InvokeOutputBinding(ctx context.Context, name, operation string, data []byte) error` is for simplicity if all you so is invoking a binding (most of the cases right now). As a new user, I wouldn't even know about InputBindings so the above method feels natural. Still, we can remove it if there is consensus

client/invoke.go
amanbha: Got it, in that case name doesn't convey what it's doing, if seems I need to give it a json from the name.

client/binding.go
amanbha: Sure, lets have consistency with methods with `JSON` int he name. Either both variations should have 2 options or they both have single variation.

client/state.go
amanbha: If SaveState takes an extra ETag it keeps it simple.

amanbha:
> @amanbha @yaron2 Given the amount of edits in this PR, and the fact that this PR doesn't actually hyde the auto generated protos that are already available in master, is there perhaps a case for focusing on cleaning up the core methods and then do small incremental improvements to the custom client after?

The auto generated sdk doesn't need to hidden, it can move to a different `autogenerated1 package. The custom sdk should not bleed any of the underlying autogenerated types. 
Sure, lets clean up the core methods as per the comments above & open bugs for remaining improvements.

client/state.go
mchmarny: Not sure I follow. SaveState takes single parameter right now `State`

```go
State struct {
    StoreName string
    States    []*StateItem
}
```

If you add Etag to the signature, what would it do? Each one of the state items has its own Etag already, no?

mchmarny:
The autogenerated (gRPC) client stubs are where they always were, `dapr/proto`, they are not headen. The value of that is that in some cases the user may want to reach around the SDK signatures to gain more control. Why would we want to move these to `autogenerated1 package? They are already versioned: `dapr/proto/common/v1` and `dapr/proto/runtime/v1`

client/binding.go
mchmarny: It seems the JSON methods were causing confusion so I removed them in the latest commit. Resolving this comments to sync with the current code state.

client/invoke.go
mchmarny: Same here, since the JSON methods were causing some confusion I removed them across the entire SDK. Users can serialize their structs themselves

amanbha:
> The autogenerated (gRPC) client stubs are where they always were, `dapr/proto`, they are not headen. The value of that is that in some cases the user may want to reach around the SDK signatures to gain more control. Why would we want to move these to `autogenerated1` package? They are already versioned: `dapr/proto/common/v1` and `dapr/proto/runtime/v1`

Thats great, if they are in separate package. this is what I was referring to. If users want to use autogenerated sdk they import the package for it, else they continue to use custom SDK.

client/state.go
amanbha: sorry, I meant SaveStateWithData

mchmarny: No patience required, this looks way better now. These massive PRs aren't easy to land

mchmarny: Added etag input param in the SaveStateWithData method

mchmarny:
@yaron2 @youngbupark Given the global changes this PR lands in the repo let's get 2+ reviewers on this before merging youngbupark

mchmarny:
@yaron2 ?

Issue #26, opened 2020-07-01
Add serving side go-sdk support


mchmarny:
Currently the go SDK supports client side only. As in app invoking Dapr. There is a lot of gRPC plumbing that could be simplify to help users create Dapr serving apps. This proposal is to create new package in the SDK called `server` to go along the current `client` package and provide that functionality. 

Example developer experience with the gRPC server (similar server implementation will be required for HTTP)

```go
import (
	"log"
	daprd "github.com/dapr/go-sdk/server/grpc"
)

func main() {
	// create a Dapr service server
	server, err := daprd.NewServer(50001)
	if err != nil {
		log.Fatalf("failed to start the server: %v", err)
	}

	// add some invocation handlers
	server.AddInvokeHandler("EchoMethod", echoHandler)

	// add some topic subscriptions
	server.AddTopicSubscription("messages", messageHandler)

	// start the server
	if err := server.Start(); err != nil {
		log.Fatalf("server error: %v", err)
	}
}

func echoHandler(contentTypeIn string, dataIn []byte) (contentTypeOut string, dataOut []byte) {
	content := string(dataIn)
	log.Printf("content: %s", content)
	return "text/plain; charset=UTF-8", []byte(content)
}

func messageHandler(event *daprd.TopicEvent) error {
	log.Printf("event - Topic:%s, ID:%s, Data: %s", event.Topic, event.ID, string(event.Data))
	return nil
}
```

mchmarny:
There are working examples of this PR to illustrate the usability 

**gRPC Serving**
https://github.com/mchmarny/dapr-grpc-service-template/blob/master/main.go

**gRPC Event Subscriber** 
https://github.com/mchmarny/dapr-grpc-event-subscriber-template

**HTTP Event Subscriber**
https://github.com/mchmarny/dapr-event-subscriber-template

mchmarny:
gRPC portion of this issue has already been [merged](https://github.com/dapr/go-sdk/pull/27). The HTTP portion of this issue is waiting for [review](https://github.com/dapr/go-sdk/pull/45)

mchmarny:
HTTP implementation for serving has been reviewed and merged. Issue completed.

PR #27, opened 2020-07-01
gRPC serving implementation addressing #26

mchmarny:
This is the gRPC implementation of the serving side outlined in #26. The updating example that goes with this PR uses this capability and does simplify the creation of gRPC services.

Usage examples here:

see `server.AddInvocationHandler("echo", echoHandler)` 
https://github.com/mchmarny/dapr-grpc-service-template/blob/master/main.go

see `server.AddTopicEventHandler(topicName, eventHandler)`
https://github.com/mchmarny/dapr-grpc-event-subscriber-template/blob/master/main.go


mchmarny:
To validate, I wrote a fully functional go template that uses this PR to create even subscription service [here](https://github.com/mchmarny/dapr-grpc-event-subscriber-template). Have to admit, this makes creating gRPC services in Dapr pretty easy have.

server/http/server.go
```diff
@@ -44,27 +44,28 @@ func (s *Server) AddTopicEventHandler(topic, route string, handler func(ctx cont
...
-	s.mux.HandleFunc(route, func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Add("Content-Type", "application/json")
-		content, err := ioutil.ReadAll(r.Body)
-		if err != nil {
-			w.WriteHeader(http.StatusBadRequest)
-			return
-		}
+	s.mux.Handle(route, optionsHandler(http.HandlerFunc(
+		func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Add("Content-Type", "application/json")
```
youngbupark, line 49: Please use predefined content-type header in net/http. `http.HeaderContentType` instead of string literal.

server/http/server.go
```diff
@@ -89,6 +83,19 @@ func (s *Server) HandleSubscriptions() error {
 	return nil
 }
 
+func optionsHandler(h http.Handler) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "OPTIONS" {
```
youngbupark, line 88: same here. i would use predefiend constant in net/http pkg. `http.MethodOptions`

server/http/server.go
```diff
@@ -89,6 +83,19 @@ func (s *Server) HandleSubscriptions() error {
...
+			w.Header().Set("Access-Control-Allow-Origin", "*")
+			w.Header().Set("Access-Control-Allow-Methods", "POST,OPTIONS")
+			w.Header().Set("Access-Control-Allow-Headers", "authorization, origin, content-type, accept")
+			w.Header().Set("Allow", "POST,OPTIONS")
```
youngbupark, line 92: All these header keys are defined in net/http. please use them.

server/grpc/server.go
```diff
@@ -0,0 +1,151 @@
...
+
+// AddInvocationHandler adds provided handler to the local collection before server start
+func (s *Server) AddInvocationHandler(method string, fn func(ctx context.Context, contentTypeIn string, dataIn []byte) (contentTypeOut string, dataOut []byte)) {
+	s.invokeHandlers[method] = fn
+}
+
+// OnInvoke gets invoked when a remote service has called the app through Dapr
+func (s *Server) OnInvoke(ctx context.Context, in *commonv1pb.InvokeRequest) (*commonv1pb.InvokeResponse, error) {
+	if val, ok := s.invokeHandlers[in.Method]; ok {
+		ct, d := val(ctx, in.ContentType, in.Data.Value)
```
youngbupark, line 63: What happens when caller application pass their own protobuf message object?

server/http/server.go
```diff
@@ -0,0 +1,102 @@
...
+		})))
+
+	return nil
+}
+
+// HandleSubscriptions creates Dapr topic subscriptions
+func (s *Server) HandleSubscriptions() error {
+	s.mux.Handle("/dapr/subscribe", http.HandlerFunc(
+		func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set("Content-Type", "application/json")
```
youngbupark, line 77: same here. please use the existing const definition in net/http

server/grpc/server.go
```diff
@@ -0,0 +1,151 @@
...
+		e := &event.BindingEvent{
+			Name:     in.Name,
+			Data:     in.Data,
+			Metadata: in.Metadata,
+		}
+		err := val(ctx, e)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error executing %s binding", in.Name)
+		}
+		return &pb.BindingEventResponse{}, nil
```
youngbupark, line 147: As you may know, BindingEventResponse can store value to statestore and send the event to multiple output binding. Do you have the plan to support it? 

On the other hand, I do not understand why we have this feature in dapr runtime. It will make user confused because they can directly call dapr api explictly. need to discuss this functionality.

server/grpc/server.go
```diff
@@ -0,0 +1,151 @@
...
+		}, nil
+	}
+	return nil, fmt.Errorf("method not implemented: %s", in.Method)
+}
+
+// START TOPIC SUB
+
+// AddTopicEventHandler adds provided topic to the list of server subscriptions
+func (s *Server) AddTopicEventHandler(topic string, fn func(ctx context.Context, event *event.TopicEvent) error) {
+	s.topicSubscriptions[topic] = fn
```
youngbupark, line 76: ListTopicSubscriptions can get the topic metadata which allows users to init pubsub component with custom options. Have you considered this metadata support?

server/http/server.go
mchmarny: net/http package does not define constants for content type (https://golang.org/pkg/net/http). Is this perhaps from diff http package?

server/http/server.go
mchmarny: net/http package does not define constants for content type (https://golang.org/pkg/net/http). Is this perhaps from diff http package?

server/http/server.go
mchmarny: same as above, https://golang.org/pkg/net/http doesn't define constants for http header keys

server/grpc/server.go
mchmarny: This is serving side, the user specifies both the content ([]bytes] and the corresponding content type. Can you provide example when the current implementation does not work?

server/http/server.go
mchmarny: same as above, https://golang.org/pkg/net/http doesn't define constants for http header keys

server/grpc/server.go
mchmarny: Not sure I follow, how would that impact the user serving code?

server/grpc/server.go
youngbupark: content_type is valid only if data is used as bytes. for protobuf message, you need to pass in.Data which includes type_url property. 

please see this: https://github.com/dapr/dapr/blob/841849a6a15b27a1dd3877b332984d1a8ddfbbdd/dapr/proto/common/v1/common.proto#L54-L62

server/http/server.go
youngbupark: they have method and status.

https://golang.org/src/net/http/method.go
https://golang.org/src/net/http/status.go

server/http/server.go
youngbupark: I have confused with fasthttp package. But I would define our own const instead of using string literal.

server/http/server.go
youngbupark: I have confused with fasthttp package. But I would define our own const instead of using string literal.

server/grpc/server.go
mchmarny: Ye, I was considering this, probably by adding additional add method (e.g. `AddTopicEventHandlerWithOptions` that would include the metadata. Users implementing simple cases won't burden this way with the additional metadata unless they need it. We can add in future release

server/grpc/server.go
youngbupark: in short, have you considered to use the properties in BindingEventResponse{} ? 

https://github.com/dapr/dapr/blob/841849a6a15b27a1dd3877b332984d1a8ddfbbdd/dapr/proto/runtime/v1/appcallback.proto#L87-L109

Spec - https://github.com/dapr/docs/blob/master/reference/api/bindings_api.md#http-response-body-optional

mchmarny: We can look into implementing something there in the future, given the required understanding of gRPC from the user, I'm wondering if that user would be even using a "convenience layer" to access Dapr. They can use the Dapr gRPC API itself

server/grpc/server.go
mchmarny: ye, we can add another method to support that use-case

server/http/server.go
mchmarny: added the use of constants

server/grpc/server.go
mchmarny: Added `DataTypeURL` support to the proto content. @youngbupark PTAL

service/http/topic.go
```diff
@@ -0,0 +1,66 @@
...
+func (s *ServiceImp) AddTopicEventHandler(topic, route string, fn func(ctx context.Context, e *TopicEvent) error) error {
+	if topic == "" {
+		return errors.New("topic name required")
+	}
+
+	if route == "" {
+		return errors.New("handler route name")
+	}
+
+	sub := &Subscription{Topic: topic, Route: route}
```
youngbupark, line 33: same here, I would allow to pass metadata to the topic list

service/grpc/service.go
```diff
@@ -0,0 +1,133 @@
...
+	return &ServiceImp{
+		listener:           lis,
+		invokeHandlers:     make(map[string]func(ctx context.Context, in *InvocationEvent) (out *Content, err error)),
+		topicSubscriptions: make(map[string]func(ctx context.Context, e *TopicEvent) error),
+		bindingHandlers:    make(map[string]func(ctx context.Context, in *BindingEvent) (out []byte, err error)),
+	}
+}
+
+// ServiceImp is the gRPC service implementation for Dapr
+type ServiceImp struct {
```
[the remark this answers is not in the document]
mchmarny: Given the interface is Service, I considered GRPCService, would that be better?

service/grpc/service.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// InvocationEvent represents the input and output of binding invocation
+type InvocationEvent struct {
+	// Data is the payload that the input bindings sent.
+	Data []byte
+	// ContentType of the Data
+	ContentType string
+	// DataTypeURL is the resource URL that uniquely identifies the type of the serialized.
+	DataTypeURL string
```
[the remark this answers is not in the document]
mchmarny: Yes, there is an actual fully function example of this in example/serving/grpc/main.go

Readme.md
```diff
@@ -170,6 +184,71 @@ secret, err = client.GetSecret(ctx, "store-name", "secret-name", opt)
...
+
+Import Dapr go `service` package:
+
+```go
+daprd "github.com/dapr/go-sdk/service/http"
+```
+
+#### Event Handling 
+
+To handle events from specific topic in HTTP, first create a Dapr serving server, add topic event handler, and start the service on specific address:
```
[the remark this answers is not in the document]
mchmarny: "consistent terminologies" to what? The proto names that the SDK is abstracting? The opposite to client is server. While I agree this is perhaps diff from what we are used to dealing with the inner-workings of Dapr but that makes sense to the user.

service/grpc/topic.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	if fn, ok := s.topicSubscriptions[in.Topic]; ok {
+		e := &TopicEvent{
+			Topic:           in.Topic,
+			Data:            in.Data,
+			DataContentType: in.DataContentType,
+			ID:              in.Id,
+			Source:          in.Source,
+			SpecVersion:     in.SpecVersion,
+			Type:            in.Type,
```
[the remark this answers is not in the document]
mchmarny: That maybe the right approach but as long as the gRPC request (`pb.TopicEventRequest` defines a strongly typed struct we should just pass all the value back to the user I think without impacting the fidelity of the request message.

service/grpc/topic.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/pkg/errors"
+
+	pb "github.com/dapr/go-sdk/dapr/proto/runtime/v1"
+)
+
+// AddTopicEventHandler appends provided event handler with topic name to the service
+func (s *ServiceImp) AddTopicEventHandler(topic string, fn func(ctx context.Context, e *TopicEvent) error) error {
```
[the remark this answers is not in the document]
mchmarny: Added metadata to the AddTopicEventHandler method

service/http/service.go
```diff
@@ -0,0 +1,121 @@
...
+	Source string `json:"source"`
+	// The content type of data value.
+	DataContentType string `json:"datacontenttype"`
+	// The content of the event.
+	Data interface{} `json:"data"`
+	// Cloud event subject
+	Subject string `json:"subject"`
+	// The pubsub topic which publisher sent to.
+	Topic string `json:"topic"`
+}
```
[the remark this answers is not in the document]
mchmarny: There was a difference in type of the data field (interface{} vs []byte) that allowed each protocol to handle it better but based on the SDK scope review we had recently the decision was to support only gRPC behind the scene so I removed all the HTTP serving implementation.

service/http/topic.go
mchmarny: same as above, HTTP serving implementation was removed now.

Readme.md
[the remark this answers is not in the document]
youngbupark: Previously, we used the term `DaprClient` server, which makes everyone confused for dapr contributor and application developer. So I recall that we switched the DaprClient to App Callback. You were in this conversation. So I used the same term in python-sdk. we can use `service` term, but I would use the `app callback` term with either `service` or `server`. The different terms in the multiple places will still make application developers and dapr contributors confused.

server/grpc/server.go
yaron2: @youngbupark it's good to bring it up.

@mchmarny we should open an issue to track this.

I don't see this being a blocker for this PR.

server/grpc/server.go
mchmarny: closing since this is outdated

server/grpc/server.go
mchmarny: closing since this is outdated

service/http/topic.go
mchmarny: http serving has been removed per common SDK design decision

service/http/service.go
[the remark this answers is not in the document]
mchmarny: http serving has been removed per common SDK design decision

service/grpc/topic.go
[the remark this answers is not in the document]
yaron2: We can consider a byte array vs object in the future, but for now I'd go with a typed struct.

If we have our own custom struct, we can enrich it with an `Extension` property of `map[string]string`.

Readme.md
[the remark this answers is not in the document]
mchmarny: I'm no too locked in on one or other just can't find any SDK that has a callback for serving like behaviour unless it's a WebHook or OAuth where both have a very specific meaning associated to `callback`. Anyhow, happy to change either way as long as we are align.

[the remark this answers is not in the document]
mchmarny: I was able to sync offline with @youngbupark and we clarified the need for clarity in the readme while supporting most appropriate approach for the service/method names.

PR #42, opened 2020-08-12
HTTP serving implementation addressing #26

mchmarny:
This is the HTTP implementation of the serving side outlined in #26. The updating example that goes with this PR uses this capability and does simplify the creation of HTTP services.


mchmarny:
Closing to cleanup the commit history

PR #45, opened 2020-08-13
HTTP serving implementation

mchmarny:
This is the HTTP implementation of the serving side outlined in #26. It refactors the gRPC and HTTP common types into `commons` package and creates single serving interface that both HTTP and gRPC implement


client/invoke.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ServiceContent the service invocation content
+type ServiceContent struct {
```
youngbupark, line 13: nit: The naming doesn't look self-descriptive :) 

Can we find the better naming ? DataContent ? ServiceInvocationContent ?

mchmarny: Agree, `DataContent` reads better and more descriptive

Issue #31, opened 2020-07-21
Implement support for Dapr API multi pub/sub in go sdk


mchmarny:
This is tracking issue in [dapr/dapr](https://github.com/dapr/dapr/issues/1709). Need to support that capability when it lands in the Dapr API. This is breaking change so critical the go-sdk capability lands at the same time.

mchmarny:
done

PR #50, opened 2020-08-17
multi pub/sub support

mchmarny:
Implements multi pub/sub support outlined in #31


service/grpc/topic.go
```diff
@@ -40,10 +50,11 @@ func (s *Server) AddTopicEventHandlerWithMetadata(topic string, m map[string]str
 // To subscribe to a topic named TopicA
```
yaron2, line 50: Is this the right comment for what this method does?

yaron2: Resolved

client/pubsub.go
```diff
@@ -8,17 +8,21 @@ import (
 )
 
 // PublishEvent pubishes data onto specific pubsub topic.
-func (c *GRPCClient) PublishEvent(ctx context.Context, topic string, in []byte) error {
+func (c *GRPCClient) PublishEvent(ctx context.Context, component, topic string, in []byte) error {
 	if topic == "" {
```
yaron2, line 12: Might want to make this the same as `AddTopicEventHandler` and be consistent with the validation messages:

```
if topic == "" {
		return fmt.Errorf("topic name required")
	}
	if component == "" {
		return fmt.Errorf("component name required")
	}
```

mchmarny: Validation messages are now consistent

client/pubsub.go
```diff
@@ -10,13 +10,13 @@ import (
...
-		return errors.New("nil topic")
+		return errors.New("topic name required")
 	}
 	if component == "" {
-		return errors.New("nil component")
+		return errors.New("component name required")
 	}
 	if in == nil {
-		return errors.New("nil in")
+		return errors.New("in data required")
```
yaron2, line 19: Data is not a requirement, empty topics are valid.

Issue #56, opened 2020-08-20
Upgrade Golang Lint to 1.30.0


mukundansundar:
Upgrade Golang Lint to 1.30.0.
Add it to build workflow.

mchmarny:
Done, closing

PR #59, opened 2020-08-26
Lint on push 

mchmarny:
Adds lint to on push action per #56


mchmarny:
This addressed https://github.com/dapr/dapr/issues/1951

Issue #29, opened 2020-07-20
Ensure consistent support for parent trace ID in client requests 


mchmarny:
Validate the ability to pass existing context with parent trace ID as an input into the client calls. Ensure the ID is propagated into the client invocation against the Dapr API

mchmarny:
This needs an outline of the test cases, for example: Context resulting from client publishing to Dapr API being used in subsequent client invocations. Also, on the serving side (callback) as well.

mchmarny:
The client already allows passing trace parent ID into each call using the context. #67 adds an explicit method to make this process easier for the user.

rainfd:
why not set WithTraceID as default behavior like withAuthToken?

PR #67, opened 2020-09-04
adds append trace parent id to ctx method

mchmarny:
The client already allows passing trace parent ID into each call using the context. This just adds an explicit method to make this process easier. This concludes the client portion of issue #29

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -17,6 +17,9 @@ import (
 const (
 	daprPortDefault    = "50001"
 	daprPortEnvVarName = "DAPR_GRPC_PORT"
+	traceparentKey     = "traceparent"
+	apiTokenKey        = "dapr-api-token"
+	apiTokenEnvVarName = "DAPR_API_TOKEN"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -72,6 +75,9 @@ type Client interface {
 	// ExecuteStateTransaction provides way to execute multiple operations on a specified store.
 	ExecuteStateTransaction(ctx context.Context, store string, meta map[string]string, ops []*StateOperation) error
 
+	// WithTraceID adds existing trace ID to the outgoign context
```
yaron2, line 78: typo: `outgoing`

client/client.go
```diff
@@ -140,11 +149,21 @@ func (c *GRPCClient) Close() {
 	}
 }
 
+// WithTraceID adds existing trace ID to the outgoign context
```
yaron2, line 152: typo: `outgoing`

yaron2 (changes requested):

client/client.go
mchmarny: corrected

yaron2 (approved):
lgtm

PR #69, opened 2020-09-04
dep cache in test/release actions

mchmarny:
the repo doesn't have many dependencies to vendor, still, the go.sum based cache cuts the test builds by ~20%


yaron2 (approved):
/lgtm

Issue #28, opened 2020-07-20
Add explicit Dapr API authentication token support


mchmarny:
All the gRPC API invocation on Dapr are already wrapped in `authContext` method so if the DAPR_API_TOKEN env var is defined, the go client will already use it to populate the outgoing context. This issue is about allowing more explicit way if propagating the auth token on client creation.

mchmarny:
done

PR #73, opened 2020-09-16
explicit Dapr API authentication token support

mchmarny:
Adds explicit Dapr API authentication token support defined in #28

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -75,8 +75,11 @@ type Client interface {
 	// ExecuteStateTransaction provides way to execute multiple operations on a specified store.
 	ExecuteStateTransaction(ctx context.Context, store string, meta map[string]string, ops []*StateOperation) error
 
-	// WithTraceID adds existing trace ID to the outgoing context
+	// WithTraceID adds existing trace ID to the outgoing context.
 	WithTraceID(ctx context.Context, id string) context.Context
+
+	// WithAuthToken sets Dapr API token on the instantiated client.
+	WithAuthToken(token string)
 
 	// Close cleans up all resources created by the client.
 	Close()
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -159,11 +169,14 @@ func (c *GRPCClient) WithTraceID(ctx context.Context, id string) context.Context
 	return metadata.NewOutgoingContext(ctx, md)
 }
 
-func authContext(ctx context.Context) context.Context {
+func (c *GRPCClient) withAuthToken(ctx context.Context) context.Context {
+	if c.authToken != "" {
+		return metadata.NewOutgoingContext(ctx, metadata.Pairs(apiTokenKey, c.authToken))
+	}
+
 	token := os.Getenv(apiTokenEnvVarName)
```
mukundansundar, line 177: here the behavior is that env var always takes priority even if the token is empty string. Wouldn't that then not reset the context api token?
Scenario:
Env var is set to token of one endpoint, client need to talk to another Dapr sidecar that does not have any authentication.

mchmarny: > here the behavior is that env var always takes priority even if the token is empty string

It isn't. Look at it closely, `if c.authToken != ""` then the `metadata.Pairs` method is using the value of `c.authToken` and returns, else continues in the method...

Explicit token set by user always take priority to enable the user to create multiple instances of the client pointing to any number/combination of Dapr instances whether they are authenticating public APIs or not.

mchmarny: thanks for reaching out @markrussinovich and pointing out the need to remove (override) the env var set value. Logic adjusted. PTAL

mukundansundar (approved):

PR #78, opened 2020-09-18
cover on push

mchmarny:


.codecov.yaml
```diff
@@ -0,0 +1,4 @@
+ignore:
+  - "**/zz_generated*.go" # generated
```
youngbupark, line 2: we do not need this here.

.codecov.yaml
```diff
@@ -0,0 +1,4 @@
...
+  - "dapr/proto"
+  - "example"
```
youngbupark, line 4: is it fine to use just directory ?
```
"dapr/proto/**"
"example/**"
```

youngbupark (changes requested):

.codecov.yaml
tcnghia: Looks like it is allowed https://docs.codecov.io/docs/ignoring-paths . I wasn't aware in the other PR.

PR #83, opened 2020-09-29
[doc] Use the latest options for port

youngbupark:
Use `dapr-http-port` and `dapr-grpc-port` instead of `port`


mchmarny (approved):
/lgmt

PR #87, opened 2020-10-07
Remove set-env usage

tcnghia:
https://nvd.nist.gov/vuln/detail/CVE-2020-15228 advises that `set-env` and `add-path` are not secure, so they will be disabled very shortly on workflows. This changes `set-env` to the recommended upgrade.


mukundansundar (approved):

mchmarny (approved):
/lgmt

PR #96, opened 2020-10-23
Adds pubsub response status in HTTP

### LINKED ISSUE #95: HTTP service implementation doesn't generate expected AppResponse
In the current HTTP serving implementation the subscriber hander response is not generating the newly added status response which results in [incorrect handling](https://github.com/dapr/dapr/blob/97cf585f8ffb54c25d12c97387e86da4f624eef9/pkg/runtime/runtime.go#L1105) in Dapr

mchmarny:
Fixes #95


yaron2 (approved):
lgtm

PR #99, opened 2020-10-29
mchmarny coverage config

mchmarny:
Splitting code coverage into client and service projects for better visibility. Also, this moves the coverage to the new default branch `main`, vs the old `master`.


yaron2 (dismissed):
lgtm

yaron2 (approved):

Issue #30, opened 2020-07-20
Pub, service invocation methods support structs without serialization


mchmarny:
In addition to the current `[]byte` as an input type, provide a way to pass raw structs

mchmarny:
Assuming this would default to JSON serialization. Any objections?

PR #104, opened 2020-11-09
Pub, service invocation methods support structs without serialization

juazasan:
This a PR asking for feedback.

I have already added a first version of a method that allows providing a Struct as an input. Wondering if it would be better the method to accept only a pointer / reference instance of a value.

I would also appreciate some advise on a better name for the method name for both the Pub and Service Invocations

On the other hand, didn't have to implement the same for service invocation yet. Working on it.

Thanks.


juazasan:
Updates PR. Now ready to merge.

go-sdk client now includes two new methods allowing Pub and Service Invocations from structs (CustomContent) without serialization).

// PublishEventfromCustomContent serializes an struct and pubishes its contents as data (JSON) onto topic in specific pubsub component.
PublishEventfromCustomContent(ctx context.Context, component, topic string, in interface{}) error

// InvokeServiceWithCustomContent invokes service with custom content (struct + content type).
InvokeServiceWithCustomContent(ctx context.Context, serviceID, method string, contentType string, content interface{}) (out []byte, err error)

juazasan:
This is addressing Issue #30 https://github.com/dapr/go-sdk/issues/30

mchmarny (approved):
/lgtm

PR #112, opened 2020-11-19
fixes #111: expose listener address on server object

### LINKED ISSUE #111: Expose http/grpc server address or allow listeners to be passed in
I am trying to use this sdk for some integration tests for dapr. The server, however, doesn't allow for the IP address of the server to be discoverable. In many build environments using hardcoded ports can lead to conflicts and discovering unused ports can lead to race conditions.

I want the http.Server or grpc.Server to expose the address that is used by the implicit listener (if the address passed in was ":0", for example. 

or

I want to be able to pass in a net.Listener to the http or grpc server so I can control the address allocation apriori.

I've tried discovering an unused port, closing the listener, and passing the address of that listener. The code is cumbersome and doesn't follow conventions used by the standard library or other projects.

daniel-garcia:
This one possible approach. The grpc version of this allows a NewServiceWithListener. This fixes the problem also. Perhaps we want a NewServerWithOptions() on both servers and use functional parameters instead of a proliferation of constructors.

### DIFF
--- a/service/grpc/service.go
+++ b/service/grpc/service.go
@@ -55,6 +55,15 @@ type topicEventHandler struct {
 	meta      map[string]string
 }
 
+// Addr returns the net.Addr of the running server. If the server
+// has not yet been started it will return nil.
+func (s *Server) Addr() net.Addr {
+	if s.listener != nil {
+		return s.listener.Addr()
+	}
+	return nil
+}
+
 // Start registers the server and starts it.
 func (s *Server) Start() error {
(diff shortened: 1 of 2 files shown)


mchmarny (commented):
The approach makes sense. Please add tests and validate with the current examples.

mchmarny:
@daniel-garcia can you update the branch and add testa please

mchmarny:
See comments on #111

PR #114, opened 2020-12-02
support metadata in state bulk get state

skyao:
update GetBulkState method: 

1. add metedata parameter
2.  change return struct to BulkStateItem

The details of this PR is discussed in dapr issue :

https://github.com/dapr/dapr/issues/2368

https://github.com/dapr/dapr/issues/2498


mchmarny (changes requested):
@skyao thanks for the PR. You can update the protos in the go SDK by running `make protos` so you can uncomment the `BulkStateItem` bits and we land a clean PR. Also, could please expand the existing test to cover the metadata bits your PR adds to state. We have test coverage thresholds in place so adding these new fields to test will ensure the PR passes.

skyao:
> @skyao thanks for the PR. You can update the protos in the go SDK by running `make protos` so you can uncomment the `BulkStateItem` bits and we land a clean PR. Also, could please expand the existing test to cover the metadata bits your PR adds to state. We have test coverage thresholds in place so adding these new fields to test will ensure the PR passes.

OK, I will try to update the protos and generated go source code.

But I meet one problem: when I run "make protos" successfully without change the protos, the generated code is big different with curernt files. I think this difference due to the version of the tools.

Since I don't find any review requirements in go-sdk repo, I tried to find in dapr repo and found this:

> Because of [etcd dependency issue](https://github.com/etcd-io/etcd/issues/11563), contributor needs to use the below verisons of tools to generate gRPC protobuf clients.

* protoc version: [v3.11.0](https://github.com/protocolbuffers/protobuf/releases/tag/v3.11.0)
* protobuf protoc-gen-go: [v1.3.2](https://github.com/golang/protobuf/releases/tag/v1.3.2)
* gRPC version: [v1.26.0](https://github.com/grpc/grpc-go/releases/tag/v1.26.0)

Should we use the save version in go-sdk?

skyao:
I update to use above version, the generated code is the same as commited in dapr/dapr repo.  But it is big different with the one in go sdk. 

And in `dapr/proto/runtime/v1/dapr.pb.go` of go-sdk, I found following information:

```go
// Code generated by protoc-gen-go. DO NOT EDIT.
// versions:
// 	protoc-gen-go v1.25.0
// 	protoc        v3.13.0
// source: dapr/proto/runtime/v1/dapr.proto
```

Compare to the version requirements in dapr/dapr:

```
protoc version: v3.11.0
protobuf protoc-gen-go: v1.3.2
gRPC version: v1.26.0
```

So to generated the code in go-sdk, which version should we use?

mchmarny:
We should follow the newer one, in dapr/dapr. Let me take a look at the setup in makefile.

mchmarny:
OK, so ye, we should use the new one. The `make protos` step in go-sdk makefile just uses whatever version of protoc you have installed locally. I just ran it and got: 

```shell
// 	protoc-gen-go v1.25.0
// 	protoc        v3.14.0
```
I know @youngbupark was going to be updating the protoc in dapr/dapr. 

Let me update the protos in a separate PR then you can re-base on that.

mchmarny:
OK, I got #116 in, one this lands you can update the metadata PR.

mchmarny:
OK, #116 landed, so you just need to resolve these few conflicts and you should be good to go.

skyao:
@mchmarny Code rebased now.

client/state.go
```diff
@@ -240,11 +250,17 @@ func (c *GRPCClient) GetBulkState(ctx context.Context, store string, keys []stri
...
 	for _, r := range results.Items {
-		item := &StateItem{
-			Key:   r.Key,
-			Etag:  r.Etag,
-			Value: r.Data,
+		item := &BulkStateItem{
+			Key: r.Key,
 		}
+		if r.Error != "" {
+			item.Error = r.Error
```
mchmarny, line 257: Default of a string field is `""` already so need for the If statement

client/state.go
```diff
@@ -240,11 +250,17 @@ func (c *GRPCClient) GetBulkState(ctx context.Context, store string, keys []stri
 	}
 
 	for _, r := range results.Items {
-		item := &StateItem{
-			Key:   r.Key,
-			Etag:  r.Etag,
-			Value: r.Data,
+		item := &BulkStateItem{
```
mchmarny, line 253: Default of a string field is "" already so need for the If statement. In fact you can just set everything in the constructor

dapr/proto/runtime/v1/dapr.pb.go
```diff
@@ -14,6 +14,7 @@ package runtime
 import (
 	v1 "github.com/dapr/go-sdk/dapr/proto/common/v1"
 	proto "github.com/golang/protobuf/proto"
+	any "github.com/golang/protobuf/ptypes/any"
```
mchmarny, line 17: Your PR should not be altering the autogenerated files.

mchmarny (changes requested):
Left a couple of comments

client/state.go
skyao: OK. Set all the fields in construtor now.

mchmarny (approved):
/lgtm

Issue #118, opened 2020-12-09
Parameterize request method verb on service invocation


ystyle:
dapr install in namespace myns, appid=myapp, and http handle router `/test`
forward port on local: 

    kubectl port-forward svc/myapp 50001:50001 3500:3500 8080:8080 -n myns

invoke in go sdk:

     res, err := client.InvokeService(ctx, "myapp.myns", "test")

this result is :  `error invoking service: rpc error: code = NotFound desc = Not Found`

browser open or call it use curl: `http://localhost:3500/v1.0/invoke/myapp.myns/method/test ` it has correct result

ps: Is there any other way to call the service locally

ystyle:
deploy 2 app in k8s. 
- app1: /test
- app2: `/test2` and use sdk invoke app1 /test 
    res, err := client.InvokeService(ctx, "myapp.myns", "test")


forward app2 port on local:

     kubectl port-forward svc/myapp2 50001:50001 3500:3500 8080:8080 -n myns

call app2/test2 use curl: 
```shell
ystyle@Archlinux  ~  curl -v 127.0.0.1:8080/test2                                                   
*   Trying 127.0.0.1:8080...
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET /test2 HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.73.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: application/json; charset=utf-8
< Set-Cookie: go_session_id=MGU1OTEzYTEtNmEzOC00NzUwLWE0YmUtMTJmZjE4ODU1ZTU2.a8b2c31a9c5286837d560cb3ac3bd85fdc6371b4; Path=/; Expires=Wed, 16 Dec 2020 03:48:01 GMT; Max-Age=604800; HttpOnly
< Date: Wed, 09 Dec 2020 03:48:01 GMT
< Content-Length: 87
< 
* Connection #0 to host 127.0.0.1 left intact
{"code":-1,"msg":"error invoking service: rpc error: code = NotFound desc = Not Found"}%                                                                          ystyle@Archlinux  ~  


```

ystyle:
go-sdk can't set  HttpExtension 
https://github.com/dapr/java-sdk/blob/master/examples/src/main/java/io/dapr/examples/invoke/http/InvokeClient.java#L35-L37

mchmarny:
[...]

PR #119, opened 2020-12-09
request method verb on service invocation

mchmarny:
Parameterizes request method verb on service invocation #118


yaron2 (approved):
lgtm

PR #126, opened 2020-12-15
adds bulk secret

mchmarny:
Implements the latest protos


client/client.go
```diff
@@ -57,6 +57,9 @@ type Client interface {
 	// GetSecret retreaves preconfigred secret from specified store using key.
 	GetSecret(ctx context.Context, storeName, key string, meta map[string]string) (data map[string]string, err error)
 
+	// GetBulkSecret retreaves all preconfigred secrets for this application.
```
yaron2, line 60: nit: retrieves

yaron2 (changes requested):

client/client.go
```diff
@@ -57,6 +57,9 @@ type Client interface {
 	// GetSecret retreaves preconfigred secret from specified store using key.
```
yaron2, line 57: retrieves

client/client.go
mchmarny: Hm, I ran make spell on this and for some reason didn't catch. On it.

client/client.go
mchmarny: Fixed all 12 of those ;)

yaron2 (approved):
LGTM

PR #127, opened 2020-12-15
feat: Invoke will extract querystring from method

rrylee:
support set querystring when invoke

### DIFF
--- a/client/invoke.go
+++ b/client/invoke.go
@@ -39,9 +39,16 @@ func (c *GRPCClient) invokeServiceWithRequest(ctx context.Context, req *pb.Invok
 	return
 }
 
-func verbToHTTPExtension(verb string) *v1.HTTPExtension {
+func queryAndVerbToHTTPExtension(query string, verb string) *v1.HTTPExtension {
+	var queryMap = map[string]string{}
+	for _, item := range strings.Split(query, "&") {
+		kv := strings.Split(item, "=")
+		if len(kv) == 2 {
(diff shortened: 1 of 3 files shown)


client/invoke.go
```diff
@@ -39,9 +39,16 @@ func (c *GRPCClient) invokeServiceWithRequest(ctx context.Context, req *pb.Invok
 	return
 }
 
-func verbToHTTPExtension(verb string) *v1.HTTPExtension {
+func queryAndVerbToHTTPExtension(query string, verb string) *v1.HTTPExtension {
```
mchmarny, line 42: Rather than passing string and splitting it this should be a map[string]string throughout

mchmarny (changes requested):
To keep it clean may be better to expose another method with the query string params as map[string]string to avoid the splits.

rrylee:
I think this code will not cause breaking change. add method will increase  learning and maintenance costs because exist three invoke method now. But it depends on you

mchmarny:
I was thinking in terms of deterministic behaviour vs inferring user intend from the string but the query string pattern is well documented so this could work too. @yaron2 what do you think?

rrylee:
our projects heavy dependency querystring and http header, and i think use sdk invoke is better than direct curl dapr v1.0 api

mchmarny:
I looked through this some more. The ability to use the existing signatures for query string requiring methods may be actually valuable. I'm going to approve it.

mchmarny (approved):
/lgtm

PR #128, opened 2021-01-05
add metadata in state get response

skyao:
In state get method, the metadata in response is missing.  

Just add it, and set the value from result.


skyao:
@mchmarny Help to review this small PR.

yaron2 (approved):
lgrm

PR #143, opened 2021-02-10
Update protobufs and query strings

### LINKED ISSUE #142: Updates for next runtime RC
Update the SDK for the following breaking changes:

1. Service invocation now handles queryparams as string and not Map of string->string. Get the latest proto and fix any code in the SDK that refers to that.

2. When content type is "application/octet-stream", CloudEvent's data will now be in `data_base64` and that should be exposed as `byte[]` to the application.

yaron2:
Closes #142


skyao:
Duplicated with my PR, I forgot this PR and submit another PR to fix the compile error yesterday.

Why this PR don't merge for so long time ( almost two month) ? What are we waiting for? 

@yaron2 @mchmarny

yaron2:
> Duplicated with my PR, I forgot this PR and submit another PR to fix the compile error yesterday. I checked the code, yaron's PR is almost the same as my PR. We can merge one of them and close both.
> 
> I have a question that why this PR don't merge for so long time ( almost two month) ? What are we waiting for? Is there anything special that I don't know?
> 
> @yaron2 @mchmarny

This was a mistake on my part, I simply forgot to merge. We need to enable a stale PR alert on this repository. Closing this as I merged yours.

PR #146, opened 2021-02-25
add publishes custom event

cinience:


mikeee:
@cinience This has been usurped by #179 which implemented the features proposed in this PR (handled by functional options)

PR #158, opened 2021-04-17
fix service invoke handler always return text/plain content type

blackstorm:
All headers should be added before w.WriteHeader(status code)


yaron2 (approved):
lgtm

PR #160, opened 2021-04-19
feat(state): SaveState expose StateOptions

1046102779:
When we use go-sdk's GetState, we don't know consistency default rules.

i use GetState method it running ok in redis test environment, but i publish server to production environment , redis configuration permissions reduced to cause bellow failed. This is a fatal problem.

```shell
(error) ERR unkown command 'WAIT' or wrong number of arguments
```

image


yaron2 (approved):
lgtm

yaron2:
Thanks for this

PR #162, opened 2021-04-27
Added Server.Unwrap function

handicraftsman:
This is a tiny commit, mostly for internal use.

I did it to support adding middlewares (such as logging, which is my case) over the http.ServeMux generated by the Dapr SDK.

It is probably not the cleanest way to do that though, as you still need to cast the `daprd.NewServiceWithMux()` output with `s.(*daprd.Server)`

### DIFF
--- a/service/http/service.go
+++ b/service/http/service.go
@@ -44,6 +44,11 @@ func (s *Server) Start() error {
 	return server.ListenAndServe()
 }
 
+// Gets the underlying serve address and http.ServeMux
+func (s *Server) Unwrap() (string, *http.ServeMux) {
+	return s.address, s.mux
+}
+
 // Stop stops previously started HTTP service
 func (s *Server) Stop() error {
 	// TODO: implement service stop


yaron2:
@skyao please review this

skyao:
```
// Gets the underlying serve address and http.ServeMux
func (s *Server) Unwrap() (string, *http.ServeMux) {
	return s.address, s.mux
}
```

This function exposures the underlying implementation details, it is realy NOT a clean way as you say. It breaks the design, can we find a better way?

handicraftsman:
I don't know, for this specific case it would be a good idea to implement adding middlewares on top of the internal header.

Or it should be possible to start treating the returned values as public, while also making sure that you cannot use the `Server` structure anymore (because messing with these values WILL break it).

yaron2:
@skyao what do you think?

skyao:
> @skyao what do you think?

I have given my comments: I understood the requirements, but I think we need find a better way, which won't  break the design.

service/http/service.go
```diff
@@ -44,6 +44,11 @@ func (s *Server) Start() error {
 	return server.ListenAndServe()
 }
 
+// Gets the underlying serve address and http.ServeMux
+func (s *Server) Unwrap() (string, *http.ServeMux) {
+	return s.address, s.mux
```
tthophan, line 49: New function `Unwrap` added. Consider adding a test case to verify its functionality.

tthophan (commented):
**Decision:** `comment` · **Confidence:** 70%

Added `Unwrap` function to `Server` type. This exposes the underlying address and `http.ServeMux`. No tests were added for this new functionality.

PR #163, opened 2021-04-28
fix PublishEventfromCustomContent missing DataContentType

### LINKED ISSUE #161: PublishEventfromCustomContent doesn't set DataContentType
**Describe the bug**
```PublishEventfromCustomContent``` serializes the input payload to JSON and stuffs it into gRPC request. However it doesn't set the ```DataContentType``` field. Dapr later will default it to "text/plain" during event consumption - this can cause confusions.

Is this intended?

**To Reproduce**
Call PublishEventfromCustomContent to publish an event and examine the data content type.

**Expected behavior**
If the payload is JSON serialized the DataContentType should be set accordingly - "application/json" in this case.

blackstorm:
#161 set `application/json` DataContentType for `PublishEventfromCustomContent`

### DIFF
--- a/client/pubsub.go
+++ b/client/pubsub.go
@@ -47,9 +47,10 @@ func (c *GRPCClient) PublishEventfromCustomContent(ctx context.Context, pubsubNa
 	}
 
 	envelop := &pb.PublishEventRequest{
-		PubsubName: pubsubName,
-		Topic:      topicName,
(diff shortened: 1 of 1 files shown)


yaron2 (approved):
lgtm

yaron2:
Thanks!

PR #173, opened 2021-06-15
Added PublishEventWithMetadata

### LINKED ISSUE #164: Add metadata support for PublishEvent API
I cannot set the partition key for Kafka without metadata support in PublishEvent API . 

Add extra argument  "metadata" of type map[string]string to this API and so that it can be set in PublishEventRequest at https://github.com/dapr/go-sdk/blob/main/client/pubsub.go#L20

ItalyPaleAle:
Fixes #164

### DIFF
--- a/client/pubsub_test.go
+++ b/client/pubsub_test.go
@@ -42,7 +42,7 @@ func TestPublishEvent(t *testing.T) {
 
 	t.Run("with metadata", func(t *testing.T) {
 		metadata := map[string]string{
-			"ttlInSeconds": "120",
+			"foo": "bar",
 		}
 		err := testClient.PublishEventWithMetadata(ctx, "messages", "test", []byte("ping"), metadata)
 		assert.Nil(t, err)
@@ -63,7 +63,7 @@ func TestPublishEvent(t *testing.T) {
 			Key2: "value2",
 		}
 		metadata := map[string]string{
-			"ttlInSeconds": "120",
+			"foo": "bar",
 		}
 		err := testClient.PublishEventfromCustomContentWithMetadata(ctx, "messages", "test", testdata, metadata)
 		assert.Nil(t, err)
--- a/client/pubsub_test.go
+++ b/client/pubsub_test.go
@@ -57,6 +57,18 @@ func TestPublishEvent(t *testing.T) {
 		assert.Nil(t, err)
 	})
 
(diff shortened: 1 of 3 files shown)


client/pubsub_test.go
```diff
@@ -40,6 +40,14 @@ func TestPublishEvent(t *testing.T) {
 		assert.NotNil(t, err)
 	})
 
+	t.Run("with metadata", func(t *testing.T) {
+		metadata := map[string]string{
+			"ttlInSeconds": "120",
```
berndverst, line 45: Could you please change the metadata to something else so people don't get the idea they should set message expiration this way?

berndverst (changes requested):

berndverst:
LGTM

If you are looking for other contributions related to this:
- Would be cool if we had an integration test that also sets up Kafka and then sets the TTL (which is what you implemented this feature for @ItalyPaleAle) using the GO SDK
- Update the Go SDK Readme to show an example that sets the TTL: https://github.com/dapr/go-sdk/#pubsub
- Update the docs at https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-message-ttl/ (this would need to go into the 1.3 branch)

berndverst (approved):

berndverst:
@yaron2 can you please review this PR also and merge if it looks good to you?

tanvigour (approved):
Looks good to me. Thanks @ItalyPaleAle !

ItalyPaleAle:
@yaron2 as I was talking to Bernd yesterday over Teams, while this PR does fix 164 perhaps there may be a better approach with what I proposed in 174. Can you take a look and let me know if I should scrap this PR? If you prefer the approach in 174, i should be able to make a new PR fairly quickly.

yaron2:
> @yaron2 as I was talking to Bernd yesterday over Teams, while this PR does fix 164 perhaps there may be a better approach with what I proposed in 174. Can you take a look and let me know if I should scrap this PR? If you prefer the approach in 174, i should be able to make a new PR fairly quickly.

Thanks for taking a look at this, I agree that this can be done in a different way. Please see my comment here: https://github.com/dapr/go-sdk/issues/174#issuecomment-864299374.

I'm closing this PR for now, we can re-open if needed.

PR #176, opened 2021-06-16
issue: hide tips

daixiang0:
Signed-off-by: Long Dai <long0dai@foxmail.com>

Tips may not be deleted, remove them from contents would make it more clear.


yaron2 (approved):
lgtm

PR #179, opened 2021-06-19
Rationalize PublishEvent* APIs

### LINKED ISSUE #174: Rationalize PublishEvent* APIs
Currently, the Go SDK has 2 methods to publish events:

- [`PublishEvent`](https://github.com/dapr/go-sdk/blob/aec809a7d5dfda7c9d5e45d81b10db17481cb1f6/client/pubsub.go#L12)
- [`PublishEventfromCustomContent`](https://github.com/dapr/go-sdk/blob/aec809a7d5dfda7c9d5e45d81b10db17481cb1f6/client/pubsub.go#L35) (also note the F should have been uppercased here)

The latter doesn't really do much more than the former, besides calling `json.Marshal` internally and setting the ContentType to `application/json`.

With #173, I've proposed adding 2 more variants for those methods: `PublishEventWithMetadata` and `PublishEventfromCustomContentWithMetadata` (F is lowercased here for consistency). These address #164 and also allow doing things like setting the TTL.

If #173 is merged (which I hope it is!) we will have 4 methods.

Yet, there will still be things that can't be done: setting options that are already available in the protobuf definition, but can't be set using the higher-level PublishEvent* APIs. One example includes setting the `ContentType` of the document (it's set to `text/plain` for the "regular" methods and `application/json` for the "fromCustomContent" methods). In the future, we could expect that there will be other things that can't be set using the higher level APIs.

I am proposing that the PublishEvent* APIs be rationalized:

1. Ideally we would have less than 4 methods
2. It should be possible to have access to a lower-level function that allows passing the protobuf object directly or a similar object (such as [`SaveBulkState`](https://github.com/dapr/go-sdk/blob/aec809a7d5dfda7c9d5e45d81b10db17481cb1f6/client/state.go#L216) for saving state)

For example, it could be:

- Deprecate PublishEventfromCustomContentWithMetadata and remove it in version 2.0
- Change PublishEvent's last parameter to be `interface{}`:
  - If the argument is a []byte or string, it will maintain the current behavior
  - If it's a protobuf object or something similar to what SaveBulkState uses, then use that object which allows setting all options.
  - If it's something else, it will be JSON-ified like PublishEventfromCustomContentWithMetadata does today (and set content type to `application/json`)

This will remain backwards-compatible today (PublishEventfromCustomContentWithMetadata can print a log message that it's deprecated but can continue to work until v2.0).

I am happy to propose a PR if you accept this suggestion, or discuss it further.

ItalyPaleAle:
Allow passing options such as Content-Type and Metadata in a flexible and future-proof way.
The method `PublishEventfromCustomContent` is now deprecated and shows a warning in the console when used.

See #174 for context and discussion

Fixes #174, #164

### DIFF
--- a/client/pubsub.go
+++ b/client/pubsub.go
@@ -3,13 +3,17 @@ package client
 import (
 	"context"
 	"encoding/json"
+	"log"
 
 	pb "github.com/dapr/go-sdk/dapr/proto/runtime/v1"
 	"github.com/pkg/errors"
 )
 
+// PublishEventOption is the type for the functional option.
+type PublishEventOption func(*pb.PublishEventRequest)
+
 // PublishEvent publishes data onto specific pubsub topic.
-func (c *GRPCClient) PublishEvent(ctx context.Context, pubsubName, topicName string, data []byte) error {
+func (c *GRPCClient) PublishEvent(ctx context.Context, pubsubName, topicName string, data interface{}, opts ...PublishEventOption) error {
 	if pubsubName == "" {
 		return errors.New("pubsubName name required")
 	}
@@ -17,13 +21,31 @@ func (c *GRPCClient) PublishEvent(ctx context.Context, pubsubName, topicName str
 		return errors.New("topic name required")
 	}
 
-	envelop := &pb.PublishEventRequest{
+	envelope := &pb.PublishEventRequest{
 		PubsubName: pubsubName,
 		Topic:      topicName,
-		Data:       data,
+	}
+	for _, opt := range opts {
+		opt(envelope)
 	}
 
-	_, err := c.protoClient.PublishEvent(c.withAuthToken(ctx), envelop)
+	if data != nil {
+		switch d := data.(type) {
+		case []byte:
+			envelope.Data = d
+		case string:
+			envelope.Data = []byte(d)
+		default:
+			var err error
+			envelope.DataContentType = "application/json"
+			envelope.Data, err = json.Marshal(d)
+			if err != nil {
+				return errors.WithMessage(err, "error serializing input struct")
+			}
+		}
+	}
+
+	_, err := c.protoClient.PublishEvent(c.withAuthToken(ctx), envelope)
 	if err != nil {
 		return errors.Wrapf(err, "error publishing event unto %s topic", topicName)
 	}
@@ -31,33 +53,30 @@ func (c *GRPCClient) PublishEvent(ctx context.Context, pubsubName, topicName str
 	return nil
 }
 
+// PublishEventWithContentType can be passed as option to PublishEvent to set an explicit Content-Type.
+func PublishEventWithContentType(contentType string) PublishEventOption {
+	return func(e *pb.PublishEventRequest) {
+		e.DataContentType = contentType
+	}
+}
+
+// PublishEventWithMetadata can be passed as option to PublishEvent to set metadata.
+func PublishEventWithMetadata(metadata map[string]string) PublishEventOption {
+	return func(e *pb.PublishEventRequest) {
+		e.Metadata = metadata
+	}
+}
+
 // PublishEventfromCustomContent serializes an struct and publishes its contents as data (JSON) onto topic in specific pubsub component.
+// DEPRECATED: This method is deprecated and will be removed in a future version of the SDK. Please use `PublishEvent` instead.
 func (c *GRPCClient) PublishEventfromCustomContent(ctx context.Context, pubsubName, topicName string, data interface{}) error {
-	if pubsubName == "" {
-		return errors.New("pubsubName name required")
-	}
-	if topicName == "" {
-		return errors.New("topic name required")
-	}
+	log.Println("DEPRECATED: client.PublishEventfromCustomContent is deprecated and will be removed in a future version of the SDK. Please use `PublishEvent` instead.")
 
-	bytes, err := json.Marshal(data)
-
+	// Perform the JSON marshaling here just in case someone passed a []byte or string as data
+	enc, err := json.Marshal(data)
 	if err != nil {
 		return errors.WithMessage(err, "error serializing input struct")
 	}
 
-	envelop := &pb.PublishEventRequest{
-		PubsubName:      pubsubName,
-		Topic:           topicName,
-		Data:            bytes,
-		DataContentType: "application/json",
-	}
-
-	_, err = c.protoClient.PublishEvent(c.withAuthToken(ctx), envelop)
-
-	if err != nil {
-		return errors.Wrapf(err, "error publishing event unto %s topic", topicName)
-	}
-
-	return nil
+	return c.PublishEvent(ctx, pubsubName, topicName, enc, PublishEventWithContentType("application/json"))
 }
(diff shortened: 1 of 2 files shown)


ItalyPaleAle:
@yaron2 Here's the PR. If you are ok with how this works, I can update the docs too.

I haven't updated the tests because... without some proper integration testing, there's not much to test anyways, as this just sets some values in a struct that a testing function wouldn't be able to check. However I can look into adding some mock methods to be able to test that the code sets the values as expected.

yaron2:
> @yaron2 Here's the PR. If you are ok with how this works, I can update the docs too.
> 
> I haven't updated the tests because... without some proper integration testing, there's not much to test anyways, as this just sets some values in a struct that a testing function wouldn't be able to check. However I can look into adding some mock methods to be able to test that the code sets the values as expected.

Thanks, I will review this week.

yaron2:
@skyao also take a look please

client/pubsub.go
```diff
@@ -3,61 +3,80 @@ package client
...
-		Data:       data,
+	}
+	for _, opt := range opts {
+		opt(envelope)
+	}
+
+	if data != nil {
+		switch d := data.(type) {
+		case []byte:
+			envelope.Data = d
```
pkedy, line 35: Does a `DataContentType` of `application/octet-stream` make sense here?

client/pubsub.go
```diff
@@ -3,61 +3,80 @@ package client
...
+func (c *GRPCClient) PublishEvent(ctx context.Context, pubsubName, topicName string, data interface{}, opts ...PublishEventOption) error {
 	if pubsubName == "" {
 		return errors.New("pubsubName name required")
 	}
 	if topicName == "" {
 		return errors.New("topic name required")
 	}
 
-	envelop := &pb.PublishEventRequest{
+	envelope := &pb.PublishEventRequest{
```
pkedy, line 24: Nit: rename `envelope` to `request` because technically it isn't a Cloud Event envelope.

pkedy (commented):
Just a couple questions re: Content Types

client/pubsub.go
ItalyPaleAle: I will rename it. I just used the old name (less the typo), but agree with you

client/pubsub.go
```diff
@@ -3,61 +3,80 @@ package client
...
+	for _, opt := range opts {
+		opt(envelope)
+	}
+
+	if data != nil {
+		switch d := data.(type) {
+		case []byte:
+			envelope.Data = d
+		case string:
+			envelope.Data = []byte(d)
```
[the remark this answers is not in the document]
ItalyPaleAle: The old method didn't set any content type. According to the Dapr API docs:

> Unless specified, Dapr assumes text/plain

So for a string type, `text/plain` should be default. However, I'm ok with making it more explicit if you'd like.

client/pubsub.go
ItalyPaleAle: This is more tricky. As mentioned, the method tries to be backwards-compatible. Previously, the method only accepted a `[]byte` value as parameter (now it's an empty interface), and it didn't specify any content-type so by default that became `text/plain` (see my other comment).

I think that changing this to `application/octet-stream` would be a backwards-incompatible change that has the potential of breaking apps silently

client/pubsub.go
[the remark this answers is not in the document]
pkedy: I'll leave it to @yaron2 and @skyao's discretion. :)

yaron2:
ping @ItalyPaleAle

ItalyPaleAle:
@yaron2 thanks for confirming the changes you wanted. I've made the last change @pkedy requested and I think this should be good to merge

yaron2:
> @yaron2 thanks for confirming the changes you wanted. I've made the last change @pkedy requested and I think this should be good to merge

Will come back to this soon.

ItalyPaleAle:
Hi is there any update on this PR?

yaron2:
@beiwei30 can you please review this PR?

client/pubsub.go
```diff
@@ -3,61 +3,80 @@ package client
...
+	if data != nil {
+		switch d := data.(type) {
+		case []byte:
+			request.Data = d
+		case string:
+			request.Data = []byte(d)
+		default:
+			var err error
+			request.DataContentType = "application/json"
+			request.Data, err = json.Marshal(d)
```
daixiang0, line 41: Need absolute assign `err` on L39?

client/pubsub.go
```diff
@@ -3,61 +3,80 @@ package client
...
+// PublishEventfromCustomContent serializes an struct and publishes its contents as data (JSON) onto topic in specific pubsub component.
+// DEPRECATED: This method is deprecated and will be removed in a future version of the SDK. Please use `PublishEvent` instead.
+func (c *GRPCClient) PublishEventfromCustomContent(ctx context.Context, pubsubName, topicName string, data interface{}) error {
+	log.Println("DEPRECATED: client.PublishEventfromCustomContent is deprecated and will be removed in a future version of the SDK. Please use `PublishEvent` instead.")
 
+	// Perform the JSON marshaling here just in case someone passed a []byte or string as data
+	enc, err := json.Marshal(data)
 	if err != nil {
-		return errors.Wrapf(err, "error publishing event unto %s topic", topicName)
+		return errors.WithMessage(err, "error serializing input struct")
```
daixiang0, line 78: LGTM overall, just need add json marshal releated tests.

client/pubsub.go
ItalyPaleAle: Yes because it needs to be defined before L41. Otherwise, `request.Data, err := json.Marshal(d)` would throw an error because it can't initialize `request.Data`

drewwells:
Any updates on this? We're eagerly waiting the arrival of metadata support in pubsub.

client/pubsub.go
yaron2: @ItalyPaleAle please fix the above and we can merge :)

ItalyPaleAle: Sorry I missed this notification.

If your ask was to add a test for this line, I've added it now.

However... This line was in previously-written code ([L46](https://github.com/dapr/go-sdk/pull/179/files#diff-7101101540524cbdbaa7b77971417897e57cb83c73d452bd214881bc22dae88cL46)) which existed before this PR It's also in a now-deprecated method. But ok :)

ItalyPaleAle:
Tests are failing but looks like it's an issue with the CI?

ItalyPaleAle:
I made a force-push but the CI continues to fail because of issues that seem to be related to the CI - something failed to download?

daixiang0:
@ItalyPaleAle like network issue, maybe force push again to re-trigger.

ItalyPaleAle:
@daixiang0 just tried again (second time) and it still fails with some CI issues.

daixiang0:
In fact there is an issue you may miss:
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.31.0-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Error: deprecatedComment: use `Deprecated: ` (note the casing) instead of `DEPRECATED: ` (gocritic)
```

ItalyPaleAle:
@daixiang0 thanks for catching that. I've fixed it locally but the CI continues to fail with some issues that seem infrastructure-related, and I am not sure how to address those. A maintainer would need to help please!

ItalyPaleAle:
@daixiang0 CI **finally** seems to work. Good to merge?

drewwells:
> @daixiang0 CI **finally** seems to work. Good to merge?

Can we get this merged? It enables a lot of things missing in go-sdk

PR #187, opened 2021-07-06
Do not wrap errors for invocations

### LINKED ISSUE #186: Invoke: gRPC errors are wrapped multiple times - cannot get the status code
**Describe the bug**

Creating a client and server in Go, I need to return a proper error status code.

My handlers look like these:

```go
daprSrv.AddServiceInvocationHandler("new", func(ctx context.Context, in *common.InvocationEvent) (out *common.Content, err error) {
	if ... {
		return nil, status.Error(codes.InvalidArgument, "Invalid content type")
	}
	return res, nil
})
```

The problem is that the errors are wrapped multiple times, both in the server and in the client. Because of that, it looks impossible to retrieve the status code.

On the client-side, the error is wrapped multiple times, including on https://github.com/dapr/go-sdk/blob/ff2b836a2e9e65bf73234ccb861541b01f4fb6a6/client/invoke.go#L29

On the server-side, `OnInvoke` method wraps the error here: https://github.com/dapr/go-sdk/blob/ab924fba7d8fbf3c797c26db57ad6d3bc1a04127/service/grpc/invoke.go#L46

It's possible to unwrap the error as wrapped on the client side with a code like:

```go
for {
	unwrapped := errors.Unwrap(err)
	if unwrapped == nil {
		break
	}
	err = unwrapped
}

grpcErr, ok := status.FromError(err)
if !ok {
	// Not a gRPC error
	return
}
```

However because the error is wrapped in the server too, it's not a gRPC error object anymore, so on the client `grpcErr` has always `code = Unknown` and I can't get the status code.

**Expected behavior**

gRPC errors in the client and server should not be wrapped so the original status code and message are preserved. Otherwise, all errors arrive with status code `Unknown` and it's not possible for the client to understand what kind of error it is.

ItalyPaleAle:
On both client and server, this does not call `errors.Wrap` on errors, so the original gRPC error object (with status code and message) is preserved.

Fixes #186

### DIFF
--- a/client/invoke.go
+++ b/client/invoke.go
@@ -26,7 +26,7 @@ func (c *GRPCClient) invokeServiceWithRequest(ctx context.Context, req *pb.Invok
 
 	resp, err := c.protoClient.InvokeService(c.withAuthToken(ctx), req)
 	if err != nil {
-		return nil, errors.Wrap(err, "error invoking service")
+		return nil, err
 	}
 
 	// allow for service to not return any value
--- a/service/grpc/invoke.go
+++ b/service/grpc/invoke.go
@@ -43,7 +43,7 @@ func (s *Server) OnInvoke(ctx context.Context, in *cpb.InvokeRequest) (*cpb.Invo
 
 		ct, er := fn(ctx, e)
 		if er != nil {
-			return nil, errors.Wrap(er, "error executing handler")
+			return nil, er
 		}
 
 		if ct == nil {


ItalyPaleAle:
Note: this doesn't stop wrapping all errors. In particular, errors for bindings, queues, etc, continue to stay wrapped as there's no user-customizable gRPC error object underneath anyways.

yaron2 (approved):
lgtm

Issue #189, opened 2021-07-20
By default, gRPC limits incoming messages to 4 MB . Expose parameter to  allow user to set the limit.


alvinhenrick:
**Describe the bug**
https://docs.microsoft.com/en-us/aspnet/core/grpc/security?view=aspnetcore-5.0

_gRPC uses per-message size limits to manage incoming and outgoing messages. By default, gRPC limits incoming messages to 4 MB. There is no limit on outgoing messages._

Saving large payload to state store works after the upgrade to DAPR 1.3 rc.1 but read fails with panic error
```
== APP == dapr client initializing for: 127.0.0.1:55325
== APP == Writing large data blob...
== APP == Saved the large data blob...
== APP == Getting data from the large data blob...
== APP == 2021/07/19 16:45:48 Writing to dapr-redis took 2.119182588s
== APP == panic: error getting state: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (82903048 vs. 4194304)
== APP == 
== APP == goroutine 1 [running]:
== APP == main.main()
== APP ==       ....../store/main_dapr_statestore.go:41 +0x5ba
== APP == exit status 2
```

**To Reproduce**

 ```
 package main

import (
	"context"
	"fmt"
	"log"
	"time"

	dapr "github.com/dapr/go-sdk/client"
)

func main() {
	client, err := dapr.NewClient()
	if err != nil {
		panic(err)
	}
	defer client.Close()
	//TODO: use the client here, see below for examples

	ctx := context.Background()
	start := time.Now()
	fmt.Println("Writing large data blob...")
	// token := []byte("hello")
	token := make([]byte, 40*1024*2024)
	store := "statestore" // defined in the component YAML

	// save state with the key key1, default options: strong, last-write
	if err := client.SaveState(ctx, store, "key1", token); err != nil {
		panic(err)
	}
	fmt.Println("Saved the large data blob...")
	elapsed := time.Since(start)
	log.Printf("Writing to dapr-redis took %s", elapsed)

	// get state for key key1
	fmt.Println("Getting data from the large data blob...")
	_, err = client.GetState(ctx, store, "key1")
	if err != nil {
		panic(err)
	}
	elapsed2 := time.Since(start)
	log.Printf("Reading from dapr-redis took %s", elapsed2)

	// fmt.Printf("data [key:%s etag:%s]: %s", item.Key, item.Etag, string(item.Value))

	// delete state for key key1
[...]

PR #197, opened 2021-08-08
Add documentation and example to deal with gRPC 4mb limit

alvinhenrick:
The approach is documented to deal with 4MB gRPC limit on client side with DAPR client.

https://github.com/dapr/go-sdk/issues/189

1. Example `examples/service/custom-grpc-client/main.go` have been added working demo.
2. The README is updated at this location `examples/service/README.md`
3. Entry in Makefile to run the example.

### DIFF
--- a/examples/service/Makefile
+++ b/examples/service/Makefile
@@ -27,6 +27,14 @@ client: ## Runs the uncompiled example client code
              --log-level debug \
              go run ./client/main.go 
 
+.PHONY: custom-grpc-client
+custom-grpc-client: ## Runs the uncompiled example custom grpc client code
+	dapr run --app-id custom-grpc-client \
+		 -d ./config \
+		 --dapr-http-max-request-size 41 \
+		 --log-level debug \
+		 go run ./custom-grpc-client/main.go
+
 .PHONY: publish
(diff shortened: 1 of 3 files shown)


alvinhenrick:
@yaron2 Any thoughts ?

yaron2:
@LaurenceLiZhixin can you please review this?

yaron2:
@alvinhenrick notice the failing test.

alvinhenrick:
> @alvinhenrick notice the failing test.

yes I think linter was failing . on my local I get lot more linter errors than during the build .I have not even made changes to that are of code ....

Can we see why the workflow did not get executed when I pushed the new changes ?

yaron2:
Running, but still failing.

alvinhenrick:
> Running, but still failing.

1. ok, I realized that there is a output string match stting validation which was failing ... How do we run that validation locally ? `validation.sh` `mm.py` 
2. Also I when a push the changes the new build does not kick off .
3. I have converted to substring match (I was logging read and write time of large payload timestamp and exact match would fail obviously) and pushed the new changes let's see if validation passes this time. 

Thank for your reviews.

yaron2 (approved):

PR #201, opened 2021-09-07
decode data_base64 cloudEvent for http service

huiyangz:
- [https://github.com/cloudevents/spec/blob/master/json-format.md](url)
Example event with `Binary`-valued data

```JSON
{
    "specversion" : "1.0",
    "type" : "com.example.someevent",
    "source" : "/mycontext",
    "id" : "B234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "application/vnd.apache.thrift.binary",
    "data_base64" : "... base64 encoded string ..."
}
```
 - The pubsub component always returned as `data_base64` from rawPayload 
```
// FromRawPayload returns a CloudEvent for a raw payload on subscriber's end.
func FromRawPayload(data []byte, topic, pubsub string) map[string]interface{} {
	// Limitations of generating the CloudEvent on the subscriber side based on raw payload:
	// - The CloudEvent ID will be random, so the same message can be redelivered as a different ID.
	// - TraceID is not useful since it is random and not from publisher side.
	// - Data is always returned as `data_base64` since we don't know the actual content type.
	return map[string]interface{}{
		IDField:              uuid.New().String(),
		......
		DataBase64Field:      base64.StdEncoding.EncodeToString(data),
	}
}
```
- The NewCloudEventsEnvelope function returns Base64 encoding for binary content type also
```
       ......
        var ceData interface{}
	ceDataField := DataField
	var err error
	if contrib_contenttype.IsJSONContentType(dataContentType) {
		err = jsoniter.Unmarshal(data, &ceData)
	} else if contrib_contenttype.IsBinaryContentType(dataContentType) {
		ceData = base64.StdEncoding.EncodeToString(data)
		ceDataField = DataBase64Field
	} else {
		ceData = string(data)
	}
        ......
```
- So, it is necessary for go-sdk to support decode data_base64 cloudEvent for http service


service/http/topic_test.go
```diff
@@ -104,3 +104,39 @@ func TestAddingInvalidEventHandlers(t *testing.T) {
...
+		}
+		if err != nil {
+			assert.NoErrorf(t, err, "error rawPayload decode")
+		}
+		return
+	}
+
+	rawData := `{
+		"datacontenttype" : "application/octet-stream",
+		"data_base64" : "eyJtZXNzYWdlIjoiaGVsbG8ifQ=="
```
daixiang0, line 124: use const string.

daixiang0:
@huiyangz could we have tests for this?

huiyangz:
> @huiyangz could we have tests for this?

As it should be.

daixiang0 (approved):
LGTM.

@yaron2 please review and merge if possible.

PR #205, opened 2021-09-24
introduces a flexible client configuration pattern

drewwells:
For highly configurable clients, it's common to use a functional
    option pattern so the client API can change overtime without
    introducing backwards breaking patterns. The new functional options
    continue to expand variables and assume defaults as before. This
    eliminates the need for multiple constructors that are all creating
    the same dapr client in slightly different ways.

    The New() function purposely returns a concrete struct instead of an
    interface. This removes an unneeded abstraction and comes from the
    principal, "accept interfaces, return structs".

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -25,93 +26,86 @@ const (
 
 var (
 	logger               = log.New(os.Stdout, "", 0)
-	_             Client = (*GRPCClient)(nil)
+	_             Client = &GRPCClient{}
 	defaultClient Client
 	doOnce        sync.Once
```
drewwells, line 31: I think the desired behavior here with the doOnce/defaultClient is to provide a default client like a `http.DefaultClient`. Of course, the user is required to initialize and create it, so the value is a little less compared to `http.DefaultClient`. If we need to support this use case, then adding `WithDefault()` functional option that reuses the hidden defaultClient will bring back this behavior.

drewwells:
Introduced a couple of common Go patterns

Return structs, accept interfaces https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8
Functional options for flexible APIs https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

drewwells:
First time committing so I need help enabling the CI to run for my submission.

drewwells:
I heard the CI is working again. How can I trigger a rebuild?

CodeMonkeyLeet:
@yaron2 Can you rerun the CI checks?

yaron2:
@drewwells please resolve the conflicts.

drewwells:
Can I get approved to run these tests on pushes? I’m not going to keep up with changes in the repo when it takes 5 days to run the checks.

daixiang0:
@drewwells please sync master branch.

drewwells:
I give up.

Issue #208, opened 2021-09-30
HTTP service stop method doesn't do anything


mnussbaum:
**Describe the bug**

The HTTP service's `Stop()` method doesn't do anything. Its body just returns `nil` and there's a `TODO` comment indicating the functionality still needs to be implemented. This means that application blocking waiting for their dapr HTTP service to shutdown will wait indefinitely.

Here's the link to the `TODO` - https://github.com/dapr/go-sdk/blob/main/service/http/service.go#L49

**To Reproduce**

```go
package main                                                  
                                                              
import (                                                      
        "fmt"                                                 
        "time"                                                
                                                              
        daprd "github.com/dapr/go-sdk/service/http"           
)                                                             
                                                              
func main() {                                                 
        service := daprd.NewService(":8080")                  
                                                              
        go func() {                                           
                if err := service.Start(); err != nil {       
                        // This should print, but it never does
                        fmt.Println("Service start completed")
                }                                             
        }()                                                   
                                                              
        service.Stop()                                        
        time.Sleep(200 * time.Millisecond)                    
}                                                             
```

**Expected behavior**

Calling an HTTP service's stop method should stop the server and cause its `Start` method to return `http.ErrServerClosed`.

dmitsh:
I also run into this issue and PR https://github.com/dapr/go-sdk/pull/210 with the fix.
Feel free to review and comment

mnussbaum:
I also opened #209 with a fix. I have test cases in that PR that might be useful in your #210 as well

dmitsh:
@mnussbaum , right, I saw your PR after sending mine.
I've closed my PR in favor of yours!

mnussbaum:
Ah gotcha, thanks!

PR #209, opened 2021-09-30
Implement HTTP service stop with a 5 sec timeout

mnussbaum:
This addresses https://github.com/dapr/go-sdk/issues/208. It implements the `Stop` method on the HTTP service which is currently a [no-op with a `TODO` comment](https://github.com/dapr/go-sdk/blob/main/service/http/service.go#L49).

It sets a 5 second timeout for the shutdown process. Open to suggestions to avoid the `time.Sleep` in the tests or for a more appropriate shutdown timeout. Thanks for taking a look!

### DIFF
--- a/service/http/service.go
+++ b/service/http/service.go
@@ -1,7 +1,9 @@
 package http
 
 import (
+	"context"
 	"net/http"
+	"time"
 
 	"github.com/dapr/go-sdk/service/common"
 )
@@ -21,7 +23,11 @@ func newServer(address string, mux *http.ServeMux) *Server {
 		mux = http.NewServeMux()
 	}
 	return &Server{
-		address:            address,
+		address: address,
+		http_server: &http.Server{
+			Addr:    address,
+			Handler: mux,
+		},
 		mux:                mux,
 		topicSubscriptions: make([]*common.Subscription, 0),
 	}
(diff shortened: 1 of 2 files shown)


alvinhenrick:
thanks @mnussbaum ... @yaron2 who can review this ?

dmitsh:
I've posted #210 independently of @mnussbaum . We have very similar implementation, so I'm ok with either one.

dmitsh (dismissed):

service/http/service.go
```diff
@@ -31,23 +37,24 @@ func newServer(address string, mux *http.ServeMux) *Server {
...
 }
 
 // Stop stops previously started HTTP service.
 func (s *Server) Stop() error {
-	// TODO: implement service stop
-	return nil
+	ctxShutDown, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer func() {
+		cancel()
+	}()
```
dmitsh, line 55: nit: defer cancel()

mnussbaum: :+1: updated

yaron2:
@mnussbaum look at the linter errors.

```
 Error: don't use underscores in Go names; struct field http_server should be httpServer (golint)
  Error: don't use underscores in Go names; var `stop_err` should be `stopErr` (golint)
  Error: don't use underscores in Go names; var `start_err` should be `startErr` (golint)
```

mnussbaum:
@yaron2 thanks for pointing those out! I've addressed the variable name lint errors.

There's still one lint error in code I touched around test helper invocation, but it looks like that issue was already present in the test file before I made my changes. Let me know if you'd like me to look into it further

yaron2 (approved):
lgtm

yaron2:
Thanks for your contribution @mnussbaum

mnussbaum:
Thanks for merging this so quickly!

dmitsh:
Thank you @mnussbaum  for your contribution and @yaron2 for quick turn-around!

Issue #212, opened 2021-10-15
Dapr client doesn't report failure when instantiated without a daprd connection


mnussbaum:
**Describe the bug**

Sometimes my applications deploy and start without the sidecar container being injected. This means I need to improve the resiliency of my dapr-sidecar-injector deployment, but regardless issues like this can occur even in the most resilient of setups.

When the dapr sidecar isn't present in my application pod my application still starts and runs as if everything is OK. However, it won't receive dapr pubsub events that are crucial to its operation. This results in a silent and hard to diagnose failure mode.

Right now instantiating a dapr client returns no error if the sidecar is missing since the client is instantiated with a non-blocking connection.

**To Reproduce**

Execute the following without daprd available. I.e. without using `dapr run` and not in a k8s pod with a dapr sidecar:

```go
package main                     
                                 
import (                         
  "fmt"                          
  "github.com/dapr/go-sdk/client"
)                                
                                 
func main() {                    
  _, err := client.NewClient()   
  if err != nil {                
    panic(err)                   
  }                              
                                 
  fmt.Println("No panic")        
}                                
```

The code will not panic and the client will be instantiated without a valid connection.

**Expected behavior**

The code should panic because dapr client initialization should return an error

PR #213, opened 2021-10-15
Make client init fail if server connection fails

mnussbaum:
This prevents client/server connectivity issues from resulting in silent failure modes and ensures they're caught as early as possible. It fixes https://github.com/dapr/go-sdk/issues/212

Moving to a blocking connection might be a larger change then I realize, if so I'm curious if there are other methods you'd suggest to test connection connectivity.

Also open to suggestions about the best connection timeout. Right now it's set to one second


yaron2:
Thanks for you contribution @mnussbaum

yaron2 (approved):

mnussbaum:
Thanks for merging!

PR #216, opened 2021-11-11
add support for state query API

dmitsh:
Adding support for state query API.

Resolves issue #215

### DIFF
--- a/client/state.go
+++ b/client/state.go
@@ -134,6 +134,21 @@ type SetStateItem struct {
 	Options  *StateOptions
 }
 
+// QueryItem represents a single query result item.
+type QueryItem struct {
+	Key   string
+	Value []byte
+	Etag  string
+	Error string
+}
+
+// QueryResponse represents a query result.
+type QueryResponse struct {
+	Results  []QueryItem
+	Token    string
+	Metadata map[string]string
+}
+
 // DeleteStateItem represents a single state to be deleted.
 type DeleteStateItem SetStateItem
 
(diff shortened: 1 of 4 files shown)


client/state.go
```diff
@@ -356,6 +371,39 @@ func (c *GRPCClient) GetStateWithConsistency(ctx context.Context, storeName, key
 	}, nil
 }
 
+// QueryState runs a query against state store.
+func (c *GRPCClient) QueryState(ctx context.Context, storeName, query string, meta map[string]string) (*QueryResponse, error) {
```
artursouza, line 375: @yaron2 how will we handle the API upgrade in SDKs? I am concerned about we changing the API in the runtime and suddenly the SDKs are not backwards compatible anymore. For sure we will need to keep the alpha and non-alpha APIs available for at least one release to make sure there is zero-downtime upgrade path.


I am considering to also name this API QueryStateAlpha in the SDK, similar to how the APIs are in the K8s SDKs we use in Dapr.

dmitsh: @yaron2 any comments on ^^ ?

yaron2: Sorry for the long delay, was mostly out for the past 17 days, back now :)

See my comment here: https://github.com/dapr/dapr/issues/3942

@artursouza since Alpha APIs can be modified, removed and added freely, breaking changes are of no concern for Alpha level APIs.
That said, on a case by case basis we can choose to either keep or remove an Alpha API version when advancing to a new version.

+1 on your suggestion to be consistent in the SDKs and call this `QueryStateAlpha`.

dmitsh: addressed

daixiang0:
Please fix conflicts then we can get it.

dmitsh:
> Please fix conflicts then we can get it.

done

PR #230, opened 2021-12-04
Decode CloudEvent Data consistently between HTTP and gRPC

### LINKED ISSUE #228: Inconsistent TopicEvent Data value between HTTP and gRPC service implementations.
When receiving a CloudEvent with `datacontenttype = application/json`, the value of `Data` has different types when using HTTP and gRPC services. For HTTP, you receive a `map[string]interface{}` and with gRPC `[]byte`.

**To Reproduce**

With a pubsub component named `pubsub` and topic named `inventory`:

```shell
curl -s http://localhost:3500/v1.0/publish/pubsub/inventory -H Content-Type:application/cloudevents+json --data @widget.json
```

Where `widget.json` contains:

```json
{
  "id": "2222",
  "type": "widget.v1",
  "source": "golang+dapr",
  "specversion": "1.0",
  "datacontenttype": "application/json",
  "data": {
    "id": "widget",
    "description": "Hello, Widget!",
    "price": 175.00,
    "widgetField": "widgets only"
  }
}
```

**Expected behavior**

`TopicEvent` could return `Data` as `map[string]interface{}` and have a new Field called `RawData` that is `[]byte`. Or perhaps it makes the most sense to always use `Data []byte` (and remove `DataBase64`) because the app ultimately decides how to deserialize the data. Regardless of how this is fixed, this will potentially break apps depending on which data type is preferred.

Additionally, a function called `Struct(target interface{})` could be added that `json.Unmarshal`s the raw bytes to a struct would be super useful. This could support JSON initially but expanded to support other formats by inspecting `datacontenttype`.

To work around this in my application for now I created this helper function:

```golang
func DecodeTopicEvent(e *common.TopicEvent, target interface{}) (err error) {
	var payload []byte
	switch v := e.Data.(type) {
	case []byte:
		payload = v
	case string:
		payload = []byte(v)
	case map[string]interface{}:
		// This is not ideal.
		payload, err = json.Marshal(v)
		if err != nil {
			return err
		}
	}
	return json.Unmarshal(payload, target)
}
```

pkedy:
This is a proposed solution #228 to gather feedback.

Closes #228


service/grpc/topic.go
```diff
@@ -62,13 +63,28 @@ func (s *Server) OnTopicEvent(ctx context.Context, in *pb.TopicEventRequest) (*p
 	}
 	key := fmt.Sprintf("%s-%s", in.PubsubName, in.Topic)
 	if h, ok := s.topicSubscriptions[key]; ok {
+		data := interface{}(in.Data)
+		if len(in.Data) > 0 {
+			switch in.DataContentType {
+			case "application/json":
```
artursouza, line 69: It can also be "application/json; UTF=8", so you probably need to normalize the input.

artursouza (changes requested):

service/grpc/topic.go
pkedy: I updated this code to use `mime.ParseMediaType(in.DataContentType)` for this. For now it only handles `application/json`. Detecting different prefixes/suffixes for JSON can come in future PRs if it makes sense.

service/grpc/topic.go
```diff
@@ -62,13 +65,38 @@ func (s *Server) OnTopicEvent(ctx context.Context, in *pb.TopicEventRequest) (*p
 	}
 	key := fmt.Sprintf("%s-%s", in.PubsubName, in.Topic)
 	if h, ok := s.topicSubscriptions[key]; ok {
+		data := interface{}(in.Data)
+		if len(in.Data) > 0 {
+			mediaType, _, err := mime.ParseMediaType(in.DataContentType)
+			if err == nil {
+				var v interface{}
+				switch mediaType {
+				case "application/json":
```
daixiang0, line 74: Better to use constant here.

daixiang0 (commented):
Please add tests for false cases if possible,

service/grpc/topic.go
pkedy: @daixiang0 I've noticed you frequently nitpick this in reviews. My general rule of thumb is that I do not create a constant for literals that are only used once and that have meaning built in (such as a media type). Numbers tend to get constants just for added meaning/clarity. In other cases, it can make the code harder to grok. So I think the Dapr project should have clear guidelines on this.

artursouza: I agree with Phil here. I don't think we should add constants for literals. For numbers, it makes more sense. I really don't like when someone creates a constant with the same name as the value it holds, for example.

pkedy: At a previous employer, my team used Uber's Go style guide. We might be able to review this and "borrow" what we like. https://github.com/uber-go/guide/blob/master/style.md

pkedy:
@daixiang0 the false tests are there.

service/grpc/topic.go
daixiang0: Agree.

daixiang0 (approved):

PR #233, opened 2021-12-14
Fix error saving actor state for readonly changes

timonwong:
When there is only readonly state operations (Get, for example), after invoking methods, the state save will error with "actor save state transactionally invocation request param operations is empty" and returns to the user with following error message:

```
rpc error: code = Internal desc = error invoke actor method: rpc error: code = Internal desc = error invoke actor method: error from actor service:
```

### DIFF
--- a/actor/state/state_async_provider.go
+++ b/actor/state/state_async_provider.go
@@ -75,6 +75,11 @@ func (d *DaprStateAsyncProvider) Apply(actorType, actorID string, changes []*Act
 			Value:         value,
 		})
 	}
+
+	if len(operations) == 0 {
+		return nil
+	}
+
 	return d.daprClient.SaveStateTransactionally(context.Background(), actorType, actorID, operations)
 }
 
--- a/actor/state/state_async_provider_test.go
+++ b/actor/state/state_async_provider_test.go
@@ -24,6 +24,35 @@ func TestDaprStateAsyncProvider_Apply(t *testing.T) {
 		args    args
 		wantErr bool
 	}{
+		{
+			name: "empty changes",
+			args: args{
+				actorType: "testActor",
+				actorID:   "test-0",
+				changes:   nil,
+			},
+			wantErr: false,
+		},
+		{
+			name: "only readonly state changes",
+			args: args{
+				actorType: "testActor",
+				actorID:   "test-0",
+				changes: []*ActorStateChange{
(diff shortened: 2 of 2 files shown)


timonwong:
It seems the CLA bot is not working?

actor/state/state_async_provider.go
```diff
@@ -75,6 +75,11 @@ func (d *DaprStateAsyncProvider) Apply(actorType, actorID string, changes []*Act
 			Value:         value,
 		})
 	}
+
+	if len(operations) == 0 {
```
daixiang0, line 79: Please add this case in tests.

timonwong: @daixiang0 It seems the tests are not running, In my local development environment with `go test -v ./actor/state` for example covers my changes:

image

timonwong:
Run with `go test -race -coverprofile=coverage.txt -covermode=atomic ./actor/state/...` gives me the coverage:

[coverage.txt](https://github.com/dapr/go-sdk/files/7723893/coverage.txt)

As you can see, it conains following line:

```
github.com/dapr/go-sdk/actor/state/state_async_provider.go:79.2,79.26 1 1
<snip>
github.com/dapr/go-sdk/actor/state/state_async_provider.go:79.26,81.3 1 1
```

Which indicates they are covered, maybe it's a codecov false negative?

timonwong:
Hello, can anyone have a look?

actor/state/state_async_provider.go
daixiang0: This test can be flaky :(

timonwong: > This test can be flaky :(

image

Personally, I strongly doubt it's a codecov issue which reports wrong line coverage (it should covers `if len(operations) == 0`  but wrongly calculated to wrong lines, which will never be executed)

I suggest to store the coverage output file as artifact, the compare to the codecov reports.

daixiang0: I agree but I am not familiar with codecov, @artursouza do you know who can make it better?

yaron2:
> Hello, can anyone have a look?

Yes, apologies. Many of us were out for holidays.

yaron2 (approved):

yaron2:
Thanks @timonwong!

PR #253, opened 2022-01-31
Fix misspelt activateActor name

jjcollinge:
As much as I like the idea of a Detective Actor 🕵️ - I'm think this should be `DeactiveActor`.


actor/manager/manager.go
```diff
@@ -120,8 +120,8 @@ func (m *DefaultActorManager) InvokeMethod(actorID, methodName string, request [
 	return rspData, actorErr.Success
 }
 
-// DetectiveActor removes actor from actor manager.
```
yaron2, line 123: DetectiveActor lol

yaron2 (approved):
lgtm

PR #266, opened 2022-03-11
Add FOSSA workflow

shubham1172:
Signed-off-by: Shubham Sharma <shubhash@microsoft.com>

This PR adds a CI step to ensure that dependencies with bad licenses are not introduced, along with a README badge.

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/4158

PR #267, opened 2022-03-11
FOSSA workflow should not run on forks

shubham1172:
Signed-off-by: Shubham Sharma <shubhash@microsoft.com>

If FOSSA workflows run on forks, it creates a new project on FOSSA. The workflow was introduced in https://github.com/dapr/go-sdk/pull/266


skyao (approved):
LGTM

PR #271, opened 2022-03-31
Add option to disable topic validation

yaron2:
As the Go-SDK is special in that it opens a gRPC server to subscribe for events, it isn't currently possible to allow for wildcard topics because of the existing `pubsub + topic` validation.

This PR adds the ability to optionally skip topic name validation for a given pub/sub.

Unblocks https://github.com/dapr/dapr/issues/4461.

cc @thielepaul


daixiang0 (approved):

daixiang0:
LGTM!

PR #272, opened 2022-04-04
Expose base grpcclient over client interface

### LINKED ISSUE #188: Expose GRPCClient.protoClient

In the `GRPCClient` struct, `protoClient` is the property that contains the lower-level gRPC client: https://github.com/dapr/go-sdk/blob/main/client/client.go#L165

This is currently a private property, not exposed publicly. There are situations where applications may benefit from being able to interact with this object directly, having access to the lower-level gRPC APIs.

Create a public getter method that returns the `protoClient` object so external objects can use it.

There are 2 alternatives for making low-level calls, but they're both sub-par:

1. Using the Dapr HTTP APIs which don't require access to the gRPC object, but these are potentially slower and require fiddling with the correct request/response structs manually
2. Creating a new gRPC connection, but it's less efficient than reusing the existing one.

thielepaul:
Signed-off-by: Paul Thiele <thielepaul@gmail.com>

Expose GrpcClient so that APIs currently not supported by the SDK (e.g. configuration) can be used with the sdk.

Close #188

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -166,6 +166,9 @@ type Client interface {
 
 	// ImplActorClientStub is to impl user defined actor client stub
 	ImplActorClientStub(actorClientStub actor.Client, opt ...config.Option)
+
+	// GrpcClient returns the base grpc client if grpc is used and nil otherwise
+	GrpcClient() pb.DaprClient
 }
 
 // NewClient instantiates Dapr client using DAPR_GRPC_PORT environment variable as port.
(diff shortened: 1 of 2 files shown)


thielepaul:
@yaron2 If modifying the interface is not an option, we could also export the field and the user could cast the return value of `NewClient()` to `GRPCClient`. I would prefer it as part of the interface though.

thielepaul:
@dapr/maintainers-go-sdk do you have any thougts regarding this PR?

daixiang0 (commented):
I am ok with this change. 

/cc @artursouza @yaron2 do you accept it?

yaron2:
> I am ok with this change.
> 
> /cc @artursouza @yaron2 do you accept it?

👍

yaron2 (approved):

PR #281, opened 2022-04-15
chore(client): update func definition to be clearer

Joshswooft:
Started playing around with dapr and had a slight issue as I followed the docs on the main website which suggested running dapr with: `dapr run --app-id myapp --dapr-http-port 3500`. 

When it came to creating the client I assumed I could use `dapr.NewClientWithPort("3500")` but upon further investigation I saw it's internally using a gRPC port. I think we should make it a bit clearer to the reader.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -194,7 +194,7 @@ func NewClient() (client Client, err error) {
 	return defaultClient, onceErr
 }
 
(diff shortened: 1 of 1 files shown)


yaron2 (approved):
lgtm

yaron2:
Thanks for your contribution!

PR #305, opened 2022-07-12
optimize: using `go mod init` to upgrade examples

1046102779:
/cc @daixiang0 @yaron2 

image

I hasn't permission to modify workflow configuration.

----------------------------

Extra doubt:

Can we abstract api `github.com/dapr/dapr/dapr/proto` into `github.com/dapr/protocol`? I think it is necessary.

In this way, each language sdk does not have to rely on the dapr core repository, and the pb protocols of all languages are stored in the protocol repository, which exists as a standard.


daixiang0:
> Can we abstract api github.com/dapr/dapr/dapr/proto into github.com/dapr/protocol? I think it is necessary.

> In this way, each language sdk does not have to rely on the dapr core repository, and the pb protocols of all languages are stored in the protocol repository, which exists as a standard.

Please create an issue in dapr core repo, I think it is good.

Makefile
```diff
@@ -7,7 +7,7 @@ all: help
 
 .PHONY: tidy
 tidy: ## Updates the go modules
-	go mod tidy -compat=1.17
+	go mod tidy
```
daixiang0, line 10: Do we need `-compat=1.18`?

1046102779: yes. I found that the dapr respository was golang1.18.

yaron2 (approved):
LGTM

yaron2:
Thanks @1046102779

Issue #259, opened 2022-02-18
panic: error creating default client: error creating connection to '127.0.0.1:50001': context deadline exceeded: context deadline exceeded


QQ2287991080:
i use the  https://docs.dapr.io/developing-applications/building-blocks/service-invocation/howto-invoke-discover-services/  go code .

error message:
dapr client initializing for: 127.0.0.1:50001
panic: error creating default client: error creating connection to '127.0.0.1:50001': context deadline exceeded: context deadline exceeded

When I use net SDK, it right.

drunkensqusih:
I ran into the same issue and went back to `v1.2.0` and it seems to have resolved it

weequan93:
it is working fine in mac m1 with v1.3.0.
Error appear after i build  with GOOS=linux and GOARCH=amd64 

after downgrade to v1.2.0 seem fine.

duffiye:
maybe can set param for start service "--dapr-grpc-port 50001 "

masters3d:
I am seeing the same error. I wonder what the issue is and why is not being addressed in next release.

pvr1:
I have the same error - Mac m1 on Dapr v1.8.1 using arm64 dapr-cli and rancher desktop 1.4.1 k8s-version:1.24.3 initialised with Dapr init -k

Please advise. Relevant logs etc incl below. 

kubectl logs app-bf84fc5c6-4cxtz daprd
daprd-container(here all seems fine in the sidecar): 
```
time="2022-07-20T21:28:32.003694128Z" level=info msg="starting Dapr Runtime -- version 1.8.1 -- commit f2583aa5cf166d83bedc129dfbbd720f100d73d2" app_id=app instance=app-bf84fc5c6-4cxtz scope=dapr.runtime type=log ver=1.8.1
time="2022-07-20T21:28:32.003719044Z" level=info msg="log level set to: debug" app_id=app instance=app-bf84fc5c6-4cxtz scope=dapr.runtime type=log ver=1.8.1
time="2022-07-20T21:28:32.003787044Z" level=info msg="metrics server started on :9090/" app_id=app instance=app-bf84fc5c6-4cxtz scope=dapr.metrics type=log ver=1.8.1
time="2022-07-20T21:28:32.003872419Z" level=info msg="Initializing the operator client (config: pipeline)" app_id=app instance=app-bf84fc5c6-4cxtz scope=dapr.runtime type=log ver=1.8.1
```

[...]

PR #310, opened 2022-08-05
fix: client default timeout

CeresCa:
Fix for https://github.com/dapr/go-sdk/issues/259

The default timeout is too short and may cause the main container to restart when the sidecar is starting slowly.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -36,6 +36,27 @@ func main() {
 }
 ```
 
+`NewClient` function has a default timeout for 5s, but you can customize this timeout by setting the environment variable `DAPR_CLIENT_TIMEOUT_SECONDS`.  
+For example: 
+```go
+package main
+
+import (
+	"os"
+	
+	dapr "github.com/dapr/go-sdk/client"
+)
+
+func main() {
+    os.Setenv("DAPR_CLIENT_TIMEOUT_SECONDS", "3")
+    client, err := dapr.NewClient()
+    if err != nil {
+        panic(err)
+    }
+    defer client.Close()
+}
+```
+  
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -234,6 +242,21 @@ func NewClientWithAddress(address string) (client Client, err error) {
...
+	if len(timeoutStr) == 0 {
+		return clientDefaultTimoutSeconds, nil
+	}
+	timeoutVar, err := strconv.Atoi(timeoutStr)
+	if err != nil {
+		return 0, err
+	}
+	if timeoutVar <= 0 {
+		timeoutVar = clientDefaultTimoutSeconds
+	}
```
mukundansundar, line 256: This behavior seems a bit inconsistent ... If the env var is set incorrectly then we are erroring out ... but when a negative value is set (incorrect value) then we are defaulting to 5 seconds. IMO these two paths should have the same behavior, either error out in both cases or default in both cases and print a warning log. 
@dapr/maintainers-go-sdk WDYT?

CeresCa: Modified, incorrect value is also erroring out now.

yaron2:
@CeresCa please add a documentation for this environment variable on the README.

Readme.md
```diff
@@ -36,6 +36,27 @@ func main() {
 }
 ```
 
+`NewClient` function has a default timeout for 5s, but you can custom this timeout by setting the environment variable `DAPR_CLIENT_TIMEOUT_SECONDS`.
```
yaron2, line 39: Change `can custom` to `can customize` please. Other than that, LGTM.

yaron2 (changes requested):

yaron2 (approved):

yaron2:
Thanks for your contribution @CeresCa

PR #311, opened 2022-08-08
Expose underlying gRPC client and server objects

### LINKED ISSUE #204: Expose underlying grpc client
I want to implement metadata support which is already available in the underlying grpc client but puzzling missing from the wrapped Client interface.

This is easy to implement by copy/pasting the PublishClient code, but I'm unable to retrieve a grpc client to use. So instead, I'll need to implement my own to call the grpc interface. 

It would be easier if Client interface supported retrieving the grpc client it uses.

Write my own client interface that just uses the grpc client directly.

ItalyPaleAle:
Allows using the gRPC server and client connection for other purposes as needed.

Fixes #204


client/client.go
```diff
@@ -297,23 +297,21 @@ type GRPCClient struct {
...
 	if c.connection != nil {
 		c.connection.Close()
+		c.connection = nil
 	}
 }
 
 // WithAuthToken sets Dapr API token on the instantiated client.
 // Allows empty string to reset token on existing client.
 func (c *GRPCClient) WithAuthToken(token string) {
-	c.mux.Lock()
```
ItalyPaleAle, line 314: This was not necessary

yaron2:
@ItalyPaleAle please resolve conflicts.

ItalyPaleAle:
@yaron2  done

yaron2 (approved):

PR #321, opened 2022-08-30
Add SaveStateWithETag() convenience function

mikeb26:
In the state API there's an existing DeleteStateWithETag() convenience
function but there does not appear to be an equivalent
SaveStateWithETag(). This commit adds SetStateWithETag() so that
consumers of the SDK don't have to employ the more verbose
SaveBulkState() when they only need to update a singular item.

Client code updating a single item with an etag prior to this commit
looks like:

```
item := &dapr.SetStateItem{
	Etag: &dapr.ETag{
		Value: "deadbeef",
	},
	Key:   "order_1",
	Value: []byte(data),
}

err := client.SaveBulkState(ctx, store, item)
```

Client code after this commit can reduce to:

```
err := client.SaveStateWithEtag(ctx, store, "order_1", []byte(data), "deadbeef")
```

Signed-off-by: Mike Brown <github@torvosoft.com>

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -90,6 +90,9 @@ type Client interface {
 	// SaveState saves the raw data into store using default state options.
 	SaveState(ctx context.Context, storeName, key string, data []byte, meta map[string]string, so ...StateOption) error
 
+	// SaveState saves the raw data into store using provided state options and etag.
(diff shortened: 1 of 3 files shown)


daixiang0:
Codes look good to me.

yaron2 (approved):

PR #327, opened 2022-10-04
Update protos

yaron2:
Signed-off-by: yaron2 <schneider.yaron@live.com>


yaron2:
Protos update, validated and tested against all samples. Force merging for release.

PR #328, opened 2022-10-04
Report SDK version to Dapr as User-Agent

ItalyPaleAle:
Following on the proposal dapr/dapr#5309, this PR changes the Go SDK to identify itself and its version in the User-Agent header.

The only caveat is that with Go dependencies being distributed as source code, we cannot bake the SDK version into what we distribute. To have the SDK report its version, then, I created a file `version/sdk-version` which needs to contain the version of the SDK that will be published. The `release-on-tag` workflow then checks if the version matches and if not, fails the pipeline.

PS: I had to update the linter to a newer version. The one being used before was very old and did not support the `go:embed` notation, with no way to get it to work.


client/client.go
```diff
@@ -266,10 +268,22 @@ func NewClientWithSocket(socket string) (client Client, err error) {
...
 	}
 	logger.Printf("dapr client initializing for: %s", socket)
-	addr := fmt.Sprintf("unix://%s", socket)
-	conn, err := grpc.Dial(addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	timeoutSeconds, err := getClientTimeoutSeconds()
+	if err != nil {
+		return nil, err
+	}
+	addr := "unix://" + socket
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds)*time.Second)
```
ItalyPaleAle, line 276: Working on this, I found a bug here. There's no need to save the context because it's only used by the `DialContext` method, and it's expired after the timeout anyways. So, I've removed saving the context in the object

client/client.go
```diff
@@ -219,36 +220,37 @@ func NewClientWithPort(port string) (client Client, err error) {
...
-	ctx, ctxCancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds)*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds)*time.Second)
 	conn, err := grpc.DialContext(
 		ctx,
 		address,
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithUserAgent("dapr-sdk-go/"+version.SDKVersion),
 		grpc.WithBlock(),
 	)
+	cancel()
```
ItalyPaleAle, line 239: Working on this, I found a bug here. There's no need to save the context because it's only used by the `DialContext` method, and it's expired after the timeout anyways. So, I've removed saving the context in the object

yaron2 (approved):

PR #333, opened 2022-11-12
fix #326 Differentiating between actor state key not found vs. unexpected error

### LINKED ISSUE #326: Differentiating between actor state key not found vs. unexpected error
It's not possible to distinguish between a key not being found and an arbitrary error related to state lookups for actors.  For example, the following code returns an error:

```go
err := myactor.GetStateManager().Get("bogus", &bytesArray)
```

The message of the error is `"get actor state result empty, with actorType: xxx, actorID: bogus, stateName yyy"`. 

I would like to then be able to write code like this:

```go
if err == ErrKeyNotFound {
    // ... do something for this specific case
} else if err != nil {
    // ... do something generic
}
```

However, the error returned in this case doesn't allow me to do this without inspecting the contents of the error message, which isn't ideal. Here is the code that generates this specific error: https://github.com/dapr/go-sdk/blob/c89770a47398058f6ea59f54e2065d7dcdaa7beb/actor/state/state_async_provider.go#L52-L54

As shown in my example, it would be great if there was some API or utility that allowed us to distinguish between the key not being found and an unexpected error for actor state lookups (it doesn't have to be my exact example).

wXwcoder:
…y or unmarshal fail
#326 
Signed-off-by: wangxw <996268132@qq.com>

### DIFF
--- a/actor/state/state_async_provider.go
+++ b/actor/state/state_async_provider.go
@@ -56,11 +56,11 @@ func (d *DaprStateAsyncProvider) Load(actorType, actorID, stateName string, repl
 		return errors.Errorf("get actor state error = %s", err.Error())
 	}
 	if len(result.Data) == 0 {
-		log.Printf("get actor state result empty, with actorType: %s, actorID: %s, stateName %s", actorType, actorID, stateName)
+		log.Errorf("get actor state result empty, with actorType: %s, actorID: %s, stateName %s", actorType, actorID, stateName)
 		return ErrStateEmpty
 	}
 	if err := d.stateSerializer.Unmarshal(result.Data, reply); err != nil {
-		log.Printf("unmarshal state data error = %s", err.Error())
+		log.Errorf("unmarshal state data error = %s", err.Error())
 		return ErrStateUnmarshal
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


daixiang0:
@wXwcoder thanks for your contribution, please update the title and description.

actor/state/state_async_provider.go
```diff
@@ -50,10 +56,12 @@ func (d *DaprStateAsyncProvider) Load(actorType, actorID, stateName string, repl
 		return errors.Errorf("get actor state error = %s", err.Error())
 	}
 	if len(result.Data) == 0 {
-		return errors.Errorf("get actor state result empty, with actorType: %s, actorID: %s, stateName %s", actorType, actorID, stateName)
+		log.Printf("get actor state result empty, with actorType: %s, actorID: %s, stateName %s", actorType, actorID, stateName)
```
daixiang0, line 59: Do we need to respect the log level here?

wXwcoder: yes，thanks，let me revise it

wXwcoder:
> @wXwcoder thanks for your contribution, please update the title and description.

done

daixiang0 (approved):

Issue #280, opened 2022-04-15
NewClient() bug if first call fails


telemac:
**Describe the bug**

I have a connection timeout problem, the default one second timeout is too short to allow the sidecar to be ready, so my first call to NewClient fails. Any subsequent call won't have a chance to make a good connection after that.

if the first call of dapr.NewClient fails, it returns the correct error, and daprClient is nil, which is ok. 
```go
daprClient, err = dapr.NewClient()
```
if I call it a second time after some delay, I will get daprClient==nil and err==nil, which is false. as daprClient is nil, it should retry a new connection and give the correct error for the new connection.

**To Reproduce**
if dapr.NewClient fails each time, you won't get an error, and you won't get a chance to retry a new connection in each iteration
```go
	var daprClient dapr.Client
        var err error

	for i := 0; i < 3; i++ {
		time.Sleep(time.Second * 3)
		daprClient, err = dapr.NewClient()
		if err != nil {
			logger.WithError(err).WithField("i", i).Warn("create dapr client loop")
		} else {
			break
		}
	}
	if err != nil {
		logger.WithError(err).Error("create dapr client")
		return
	}
```

**Expected behavior**

if NewClient fails, it should retry to connect on next call and return a correct error, and not use doOnce.Do inconditionally.

munjalpatel:
I am also having the same exact issue. Just can't get it to work after spending days debugging it.
Both client and server are on `v1.7.0`

@telemac were you able to make any progress?

telemac:
Hi @munjalpatel, a simple sleep for 5 seconds before calling dapr.NewClient solves the issue for now.

A dapr.NewClient taking a timeout or may be a context (the most idiomatic way) would be great too.

munjalpatel:
Well, for me, the `client` is `nil` even after sleep. Both `err` and `client` are `nil`.

kayvonkhosrowpour:
Still having the same issue.

wXwcoder:
retry to create client，can use dapr.NewClientWithPort

spolab:
Just an observation: I would expect a client to exit(1) or panic if the client cannot be created. In most container-based environments, this would trigger a respawn of the container, and the next creation of the client (the first in a new pod instance) would then work. Not sure if I would consider this a bug.

PR #341, opened 2022-12-10
Fix create new client error

Taction:
use lock instead of once to create the default client, and return the correct error on repeated calls.

The issue this pr will close:#280


yaron2:
@Taction please resolve the conflict.

yaron2 (approved):

PR #346, opened 2023-01-05
Various fixes and cleanups

ItalyPaleAle:
1. Fixed SDK not compiling and tests failing
2. Run CI tests on both Go 1.18 and 1.19
3. Increased minimum Go version to 1.18
4. Fixed "validate-examples" Action that was failing due to changes with Helm
6. Removed `github.com/pkg/errors` and `go.uber.org/atomic` - replaced with the standard library instead (the old packages have been added to the "depguard" linter)
7. Upgraded all dependencies
8. Upgraded golangci-lint to support Go 1.18+
9. Upgraded (most, but not all :( ) deprecated Actions to stop warnings from GH Actions


yaron2:
Great PR, thanks

yaron2 (approved):

PR #356, opened 2023-02-01
V1.10 Update Protos & Go Modules

robertojrojas:
V1.10 Update Protos & Go Modules


examples/hello-world/go.mod
```diff
@@ -2,25 +2,24 @@ module github.com/dapr/go-sdk/examples/hello-world
 
 go 1.17
 
+// Needed to validate SDK changes in CI/CD
+replace github.com/dapr/go-sdk => ../../
```
yaron2, line 6: Is this needed?

robertojrojas: I believe so. I don't know any other way to make sure imports like the following point to the version of code being tested locally:
```go
import (
  dapr "github.com/dapr/go-sdk/client"
)
```
If we were to remove the replace, `go mod tidy` uses the latest version of the module that exists in Github, currently for `github.com/dapr/go-sdk` is `v1.6.0`.

PR #358, opened 2023-02-13
Create `xxxContext` functions for ones which are missing a caller context.

JoshVanL:
PR adds `xxxContext` variations of functions that are currently using a `context.Background()`. This enables callers to pass their own context.

For backwards compat, no existing functions have changed their signature and behave the same as before- they are now all a wrapper for their `xxxContext` counterparts.


yaron2:
@daixiang0 please help review.

actor/actor.go
```diff
@@ -78,25 +81,40 @@ func (b *ServerImplBase) SetID(id string) {
...
 	}
 	return nil
 }
 
+//nolint:interfacebloat
 type StateManager interface {
 	// Add is to add new state store with @stateName and @value
 	Add(stateName string, value interface{}) error
+	// Add is the same as Add, but with context.
+	AddContext(ctx context.Context, stateName string, value interface{}) error
```
daixiang0, line 99: I would like `AddWithContext()` to include `Add()`, or make `AddContext()`  only to do one thing -- add context.

artursouza: Would it make more sense to deprecate this interface and create a new one? The existing one would still be supported but fallback to calling the new implementation where context is always present.

daixiang0: I agree.

@JoshVanL please update as @artursouza said.

actor/actor.go
```diff
@@ -39,64 +36,176 @@ type Server interface {
...
 	id           string
+	lock         sync.RWMutex
 }
 
+// Deprecated: Use ServerImplBaseCtx instead.
 func (b *ServerImplBase) SetStateManager(stateManager StateManager) {
+	b.ctx.lock.Lock()
+	defer b.ctx.lock.Unlock()
 	b.stateManager = stateManager
+	b.ctx.stateManager = stateManager.WithContext()
```
artursouza, line 95: Does this create a race condition where two contexts can be set while the invocation did not take place yet? Is there a lock somewhere surrounding WithContext() and SaveState()?

The lock here will avoid another race condition instead.

actor/actor.go
```diff
@@ -39,64 +36,176 @@ type Server interface {
...
 	Contains(stateName string) (bool, error)
 	// Save is to saves the state cache of this actor instance to state store component by calling api of daprd.
 	Save() error
-	// Flush is called by stateManager after Save
+	// Flush is called by StateManager after Save
 	Flush()
+
+	// Returns a new StateManagerContext with the same state as this StateManager
+	// but uses context.
+	WithContext() StateManagerContext
```
artursouza, line 193: Same question about race condition.

artursouza (changes requested):

actor/actor.go
```diff
@@ -39,64 +36,176 @@ type Server interface {
...
 	}
+
 	return nil
 }
 
+// Deprecated: StateManager is deprecated in favour of StateManagerContext.
 type StateManager interface {
 	// Add is to add new state store with @stateName and @value
-	Add(stateName string, value interface{}) error
+	Add(stateName string, value any) error
```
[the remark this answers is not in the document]
JoshVanL: No, `any` is a straight type alias to `interface{}`.

actor/actor.go
JoshVanL: Thanks, there can be. I'll add another lock to `ServerImpBase`.

JoshVanL:
Thanks @artursouza, PTAL

actor/actor.go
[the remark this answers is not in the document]
daixiang0: let's keep it as before.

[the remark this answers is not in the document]
JoshVanL: @daixiang0 any particular reason for this? The `any` keyword seems cleaner, is far more preferred in upstream go.

```bash
$ grep -inr /nix/store/ngvd1babcfc84a2rsdsvbrlmc6w3pg8m-go-1.20.1/share/go/src --exclude-dir=internal -e "any" | wc -l
8564
```

```bash
$ grep -inr /nix/store/ngvd1babcfc84a2rsdsvbrlmc6w3pg8m-go-1.20.1/share/go/src --exclude-dir=internal -e "interface{}" | wc -l
447
```

artursouza (approved):

PR #387, opened 2023-04-05
[Crypto] Implement support for high-level crypto APIs

ItalyPaleAle:
This implements support for the high-level crypto APIs implemented here: dapr/dapr#6197 

These allow encrypting and decrypting files of arbitrary length (up to 256TB :) ), while working on a stream of data.

The gRPC APIs are quite complex as they require data to be chunked, but the SDK takes care of all that complexity and offers a simple API with `Encrypt` and `Decrypt`. Both accept a readable stream as input (`io.Reader`) and return a readable stream (`io.Reader`) that receives the processed data.

Example usage. This reads a file and encrypts it, storing the result in another file:

```go
// Input file, clear-text
rf, err := os.Open("input")
if err != nil {
	panic(err)
}
defer rf.Close()

// Output file, encrypted
wf, err := os.Create("output.enc")
if err != nil {
	panic(err)
}
defer wf.Close()

// Encrypt the data using Dapr
out, err := sdkClient.Encrypt(context.Background(), rf, dapr.EncryptOptions{
	// These are the 3 required parameters
	ComponentName: "mycryptocomponent",
	KeyName:        "mykey",
	Algorithm:     "RSA",
})
if err != nil {
	panic(err)
}

// Read the stream and copy it to the out file
n, err := io.Copy(wf, out)
if err != nil {
	panic(err)
}
fmt.Println("Written", n, "bytes")
```

Decrypting works very similarly:

```go
out, err := sdkClient.Decrypt(context.Background(), rf, dapr.EncryptOptions{
	// Only required option is the component name
	ComponentName: "mycryptocomponent",
})
```


Makefile
```diff
@@ -41,8 +41,8 @@ tag: ## Creates release tag
 .PHONY: clean
 clean: ## Cleans go and generated files in ./dapr/proto/
 	go clean
-	rm -fr ./dapr/proto/common/v1/*
-	rm -fr ./dapr/proto/runtime/v1/*
+	rm -fr ./dapr/proto/common/v1/*.pb.go
```
ItalyPaleAle, line 44: This is needed because I've added a non-generated file in those folders, with code that has to be crafted by hand.

yaron2:
@daixiang0 can you please review this?

Makefile
daixiang0: no new file under common, do I miss it?

client/crypto.go
```diff
@@ -0,0 +1,272 @@
...
+	runtimev1pb "github.com/dapr/go-sdk/dapr/proto/runtime/v1"
+)
+
+// Encrypt data read from a stream, returning a readable stream that receives the encrypted data.
+// This method returns an error if the initial call fails. Errors performed during the encryption are received by the out stream.
+func (c *GRPCClient) Encrypt(ctx context.Context, in io.Reader, opts EncryptOptions) (io.Reader, error) {
+	// Ensure required options are present
+	// This short-circuits and avoids a call to the runtime
+	if opts.ComponentName == "" {
+		return nil, errors.New("option 'ComponentName' is required")
```
daixiang0, line 35: Better to make them as const.

ItalyPaleAle: I am not a fan of making errors as constants:

1. They are only used once or twice. Defining them as constants requires the person reading the code to jump back and forth to read what the constant is. It makes reading the code harder
2. The difference would be if we expected callers to perform `errors.Is` to check if the error is the same. These errors are generally “dev-time” errors so users are not expected to check if the errors match.

dapr/proto/runtime/v1/dapr_additional.go
```diff
@@ -0,0 +1,104 @@
...
+
+	commonv1pb "github.com/dapr/go-sdk/dapr/proto/common/v1"
+)
+
+// CryptoRequests is an interface for EncryptRequest and DecryptRequest.
+type CryptoRequests interface {
+	proto.Message
+
+	// SetPayload sets the payload.
+	SetPayload(payload *commonv1pb.StreamPayload)
```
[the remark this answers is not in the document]
ItalyPaleAle: Having these interfaces allow me to implement `performCryptoOperation` just once for both encrypt and decrypt. Otherwise the code has to be implemented twice.

client/bufpool.go
```diff
@@ -0,0 +1,31 @@
+/*
+Copyright 2021 The Dapr Authors
```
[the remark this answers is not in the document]
ItalyPaleAle: Changed

yaron2 (approved):

Issue #237, opened 2021-12-29
Grpc Service Start with provided grpc server


riza-wartek:
**Is your feature request related to a problem? Please describe.**
I need my app to expose both grpc service and subscribing events. Looking here and there, I came to a conclusion that I cannot do both because [`Start()`](https://github.com/dapr/go-sdk/blob/1673ee9e7aa93c98e56d7554760ed52fbcbdbd98/service/grpc/service.go#L80)  always create new grpc server

**Describe the solution you'd like**
have a new function like `StartWithServer` function with `*grpc.Server` as parameter 

**Describe alternatives you've considered**
Implement interface [described here](https://docs.dapr.io/developing-applications/integrations/grpc-integration/) and aware of all interfaces available even if the intention is to use state store and pubsub only

PR #400, opened 2023-05-19
Allow setting own gRPC server when calling `NewService`

pimmerks:
We are building an application that exposes it's own gRPC service, it also needs to subscribe to the dapr pub/sub component.

Currently we are allowed to set the listener, but this is not enough, as we found that about half of the time, our own gRPC server responds, and the other half of the time the dapr service responds with an `unknown service xxx.xxx`.

If this is merged, the user is allowed to set their own grpcServer with the `NewServiceWithGrpcServer()` func.

In action:
```golang
lis, _ := net.Listen("tcp", "localhost:8080")
server := grpc.NewServer()

// Register dapr
dapr.NewServiceWithGrpcServer(lis, server)

// Setup dapr subscriptions here

// Register own services
proto.RegisterOwnServer1(server, ownServer1{})
proto.RegisterOwnServer2(server, ownServer2{})

go server.Serve(lis)
```

In issue https://github.com/dapr/go-sdk/issues/237 the same problem has been found.

### DIFF
--- a/service/grpc/service.go
+++ b/service/grpc/service.go
@@ -39,16 +39,21 @@ func NewService(address string) (s common.Service, err error) {
 		err = fmt.Errorf("failed to TCP listen on %s: %w", address, err)
 		return
 	}
-	s = newService(lis)
+	s = newService(lis, nil)
 	return
 }
 
 // NewServiceWithListener creates new Service with specific listener.
 func NewServiceWithListener(lis net.Listener, opts ...grpc.ServerOption) common.Service {
-	return newService(lis, opts...)
+	return newService(lis, nil, opts...)
 }
 
-func newService(lis net.Listener, opts ...grpc.ServerOption) *Server {
+// NewServiceWithGrpcServer creates a new Service with specific listener and grpcServer
+func NewServiceWithGrpcServer(lis net.Listener, server *grpc.Server) common.Service {
+	return newService(lis, server)
+}
+
+func newService(lis net.Listener, grpcServer *grpc.Server, opts ...grpc.ServerOption) *Server {
 	s := &Server{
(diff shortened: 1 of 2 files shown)


service/grpc/service.go
```diff
@@ -57,7 +62,14 @@ func newService(lis net.Listener, opts ...grpc.ServerOption) *Server {
...
 	}
 
-	gs := grpc.NewServer(opts...)
+	var gs *grpc.Server
+
+	if grpcServer != nil {
+		gs = grpcServer
+	} else {
+		gs = grpc.NewServer(opts...)
+	}
```
JoshVanL, line 71: Prob don't really need a new variable here.

```suggestion
	if grpcServer == nil {
		grpcServer = grpc.NewServer(opts...)
	}
```

service/grpc/service.go
```diff
@@ -57,7 +62,14 @@ func newService(lis net.Listener, opts ...grpc.ServerOption) *Server {
...
+
 	pb.RegisterAppCallbackServer(gs, s)
 	pb.RegisterAppCallbackHealthCheckServer(gs, s)
 	s.grpcServer = gs
```
JoshVanL, line 75:
```suggestion

	pb.RegisterAppCallbackServer(grpcServer, s)
	pb.RegisterAppCallbackHealthCheckServer(grpcServer, s)
	s.grpcServer = grpcServer
```

JoshVanL (changes requested):
Thanks @pimmerks, this seems like a sensible extension to me :slightly_smiling_face:

JoshVanL (approved):

yaron2 (approved):

Issue #411, opened 2023-06-05
Panic when call server.Stop() twice


hunter007:
**I stop grpc.Service twice, and panic occurred.**

When I read service's unit test codes, I find the question:

```go
func TestServer(t *testing.T) {
	server := getTestServer()
	startTestServer(server)
	stopTestServer(t, server)
}

func startTestServer(server *Server) {
	go func() {
		if err := server.Start(); err != nil && err.Error() != "closed" {
			panic(err)
		}
	}()
}

func stopTestServer(t *testing.T, server *Server) {
	t.Helper()

	assert.NotNil(t, server)
	err := server.Stop()
	assert.Nilf(t, err, "error stopping server")
}
```

This test is invalid because server hasn't started yet when stopping it.

**To Reproduce**
Add `fmt.Println("Started")` before returing,like this:
```go
func (s *Server) Start() error {
	if !atomic.CompareAndSwapUint32(&s.started, 0, 1) {
		return errors.New("a gRPC server can only be started once")
	}
	fmt.Println("Started")
	return s.grpcServer.Serve(s.listener)
}
```

And add code `fmt.Println("Stopped")` after `err := server.Stop()` in `func stopTestServer(t *testing.T, server *Server)`, 
like this:
```go
func stopTestServer(t *testing.T, server *Server) {
	t.Helper()

	assert.NotNil(t, server)
	err := server.Stop()
	fmt.Println("Stopped")
	assert.Nilf(t, err, "error stopping server")
}
```

Run this test again, output:
```
Running tool: ./go1.20/bin/go test -timeout 30s -run ^TestServer$ github.com/dapr/go-sdk/service/grpc -v

=== RUN   TestServer
Stopped
--- PASS: TestServer (0.00s)
PASS
Started
ok  	github.com/dapr/go-sdk/service/grpc	0.140s
```

This is an error test. To correcte it, add `time.Sleep(time.Second)` to last line of `startTestServer` to ensure server has started:
```go
func startTestServer(server *Server) {
	go func() {
		if err := server.Start(); err != nil && err.Error() != "closed" {
			panic(err)
		}
	}()
	time.Sleep(time.Second)
}
```
Run this test again, output:
```
Running tool: ./go1.20/bin/go test -timeout 30s -run ^TestServer$ github.com/dapr/go-sdk/service/grpc -v

=== RUN   TestServer
Started
Stopped
--- PASS: TestServer (1.00s)
PASS
[...]

PR #412, opened 2023-06-05
Ensure that calling grpcServer.Stop() only once

hunter007:
https://github.com/dapr/go-sdk/issues/411

### DIFF
--- a/service/grpc/service.go
+++ b/service/grpc/service.go
@@ -85,6 +85,7 @@ type Server struct {
 	authToken          string
 	grpcServer         *grpc.Server
 	started            uint32
+	stopped            uint32
 }
 
 // Deprecated: Use RegisterActorImplFactoryContext instead.
@@ -101,26 +102,37 @@ func (s *Server) Start() error {
 	if !atomic.CompareAndSwapUint32(&s.started, 0, 1) {
 		return errors.New("a gRPC server can only be started once")
 	}
+
+	if atomic.LoadUint32(&s.stopped) == 1 {
+		return errors.New("the gRPC server has stopped")
+	}
+
 	return s.grpcServer.Serve(s.listener)
 }
 
 // Stop stops the previously-started service.
 func (s *Server) Stop() error {
-	if atomic.LoadUint32(&s.started) == 0 {
-		return nil
-	}
-	s.grpcServer.Stop()
-	s.grpcServer = nil
-	return nil
+	return s.gracefulStop(false)
 }
 
 // GrecefulStop stops the previously-started service gracefully.
 func (s *Server) GracefulStop() error {
+	return s.gracefulStop(true)
+}
+
+func (s *Server) gracefulStop(graceful bool) error {
 	if atomic.LoadUint32(&s.started) == 0 {
 		return nil
 	}
-	s.grpcServer.GracefulStop()
-	s.grpcServer = nil
+
+	if atomic.CompareAndSwapUint32(&s.stopped, 0, 1) {
+		if graceful {
+			s.grpcServer.GracefulStop()
+		} else {
+			s.grpcServer.Stop()
+		}
+		s.grpcServer = nil
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


service/grpc/service.go
```diff
@@ -101,26 +102,29 @@ func (s *Server) Start() error {
 	if !atomic.CompareAndSwapUint32(&s.started, 0, 1) {
 		return errors.New("a gRPC server can only be started once")
 	}
+
 	return s.grpcServer.Serve(s.listener)
 }
 
 // Stop stops the previously-started service.
 func (s *Server) Stop() error {
-	if atomic.LoadUint32(&s.started) == 0 {
```
daixiang0, line 109: If not started, double call will return nil. Is it an issue that we miss update the start status?

hunter007: Yes. I will correct it

daixiang0: Also, we can use start status to check stop, no need to add a new one.

hunter007: We can prevent restart and re-stop using `stopped`.

If no `stopped`, when `started = 0`, we can call `Start()` which set `started=1`.

But, call `Stop()`, if set `started = 0`, we can restart server? if keep `started = 1`, we can re-stop?

daixiang0: When `started` is:

- true, can stop, start/re-start but do nothing
- false, can start/re-start, stop/re-stop but do nothing

hunter007: I write a test:

```go
func TestReStopServer(t *testing.T) {
	server := getTestServer()
	startTestServer(server)
	time.Sleep(time.Second)

	assert.NotNil(t, server)
	err := server.Stop()
	assert.Nilf(t, err, "error stopping server")

	fmt.Print("Re-stop server\n")

	err = server.Stop()
	assert.Nilf(t, err, "error re-stopping server")
}
```

the test panic:

```
go test -timeout 30s -run ^TestReStopServer$ github.com/dapr/go-sdk/service/grpc -v

=== RUN   TestReStopServer
Re-stop server
--- FAIL: TestReStopServer (1.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x180 pc=0x10475103c]

goroutine 34 [running]:
testing.tRunner.func1.2({0x1048f8860, 0x104d0c280})
	/go1.20/src/testing/testing.go:1526 +0x1c8
testing.tRunner.func1()
	/go1.20/src/testing/testing.go:1529 +0x364
panic({0x1048f8860, 0x104d0c280})
	/go1.20/src/runtime/panic.go:884 +0x1f4
google.golang.org/grpc.(*Server).Stop(0x0)
	/gopkg/mod/google.golang.org/grpc@v1.55.0/server.go:1800 +0x2c
github.com/dapr/go-sdk/service/grpc.(*Server).Stop(...)
	/dapr_repos/go-sdk/service/grpc/service.go:112
github.com/dapr/go-sdk/service/grpc.TestReStopServer(0x0?)
	/dapr_repos/go-sdk/service/grpc/service_test.go:79 +0x1c4
testing.tRunner(0x140001171e0, 0x10499b2f0)
	/go1.20/src/testing/testing.go:1576 +0x104
created by testing.(*T).Run
	/go1.20/src/testing/testing.go:1629 +0x370
```

I made this PR to resolve this issue:  https://github.com/dapr/go-sdk/issues/411

service/grpc/service_test.go
mikeee, line 1: I think we should be using the `require` package for error evaluations e.g. require.Error / require.NoError

service/grpc/service_test.go
```diff
@@ -53,6 +54,7 @@ func startTestServer(server *Server) {
 			panic(err)
 		}
 	}()
+	time.Sleep(time.Second)
```
tthophan, line 57: Using `time.Sleep` in tests is flaky. Consider using a channel or `sync.WaitGroup` to signal server readiness instead.

tthophan (approved):

PR #420, opened 2023-06-15
Add support for HTTPS and H2C app protocols

ItalyPaleAle:
This PR adds support to create a `service` that uses HTTPS (incl. HTTP/2) or H2C (HTTP/2 Cleartext). Using HTTP/2 should offer significant performance improvements, for example due to support for multiplexing.

1. New method `http.NewServiceWithOptions` which allows creating a service with options, including configuring the protocol explicitly ("https", "http", "h2c") and passing TLS certificates
2. If the env var `APP_PROTOCOL` (see dapr/dapr#6512 and dapr/cli#1318) is present, uses that to detect HTTPS or H2C usage
3. If app protocol is HTTPS and no TLS certificates are passed, generates a self-signed cert automatically (they have validity of 1 year, but the expiration doesn't matter because Dapr doesn't validate TLS certificates)


daixiang0:
wait for related PRs merged.

ItalyPaleAle:
> wait for related PRs merged.

Which PR? The runtime features are merged

yaron2:
validation isn't passing.

ItalyPaleAle:
Can you re-trigger it please? It's been a while since it ran

yaron2:
Doesn't seem to be a way to re-trigger.. i'll reopen.

ItalyPaleAle:
Uhm I don't understand what's causing the tests to fail. Everything is working fine, but the failure is here:

https://github.com/ItalyPaleAle/dapr-go-sdk/blob/c449ceace2cc2a7520e411bb1ad04531d9b755f9/examples/service/README.md#L167

Which is trying to kill a process and it fails, after having already stopped Dapr?

```
(lsof -i:8080 | grep main) | awk '{print $2}' | xargs  kill
```

service/http/service.go
```diff
@@ -28,30 +33,116 @@ import (
...
-	return &Server{
-		address: address,
-		httpServer: &http.Server{ //nolint:gosec
-			Addr:    address,
-			Handler: router,
-		},
-		mux:            router,
+	if opts.Protocol == "" {
+		// If there's no user-defined protocol, try reading from the APP_PROTOCOL env var
+		opts.Protocol = common.ServiceProtocol(os.Getenv(common.AppProtocolEnvVar))
```
tthophan, line 105: Consider using `http.NoBody` or a more explicit timeout configuration if needed, but `30 * time.Second` is acceptable for `ReadHeaderTimeout` as a baseline.

tthophan (approved):

PR #424, opened 2023-07-03
Fixed: do not require a Content-Length header

ItalyPaleAle:
The HTTP service in multiple places expected a Content-Length header or it would refuse to parse the body.

This breaks streaming (see [test failures](https://github.com/dapr/dapr/actions/runs/5446419269/jobs/9907127023)), since in case the data is streamed it's not possible to know the content's byte length in advance. The Content-Length header is not required per the HTTP specs, so we should not enforce its presence.

This fixes the issue by not relying on the Content-Length header's presence.

### DIFF
--- a/service/http/topic.go
+++ b/service/http/topic.go
@@ -67,6 +67,52 @@ type topicEventJSON struct {
 	PubsubName string `json:"pubsubname"`
 }
 
+func (in topicEventJSON) GetData() (data any, rawData []byte) {
+	var (
+		err error
+		v   any
+	)
+	if len(in.Data) > 0 {
+		rawData = []byte(in.Data)
+		data = rawData
+		// We can assume that rawData is valid JSON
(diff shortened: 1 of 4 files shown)


service/http/topic.go
```diff
@@ -67,6 +67,52 @@ type topicEventJSON struct {
 	PubsubName string `json:"pubsubname"`
 }
 
+func (in topicEventJSON) GetData() (data any, rawData []byte) {
```
ItalyPaleAle, line 70: Why this change, where I split this out in a separate function, you may ask?

Well, ask the linter and its complaining about "cyclomatic complexity"...

artursouza: Can it be a private method instead?

ItalyPaleAle: Sure, done

artursouza (approved):

PR #431, opened 2023-07-06
[docs] Add query state and actors examples

hhunter-ms:
Attempt at adding examples for query state and actors to the client doc

Related issue: dapr/docs#2770


daixiang0 (approved):
LGTM

PR #437, opened 2023-08-16
Adds new `NewClientWithAddressContextNoEnvTimeout` func

JoshVanL:
As an app developer I do not want to have a default of 5 seconds timeout for connecting to daprd, and don't wish to use environment variables to configure this value. Instead, I would like to solely rely on the context I pass to the client New func.

Adds new `NewClientWithAddressContextNoEnvTimeout` func to create clientwhich ignores the `DAPR_CLIENT_TIMEOUT_SECONDS` env var.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -253,16 +253,24 @@ func NewClientWithAddress(address string) (client Client, err error) {
 // NewClientWithAddressContext instantiates Dapr using specific address (including port).
 // Uses the provided context to create the connection.
 func NewClientWithAddressContext(ctx context.Context, address string) (client Client, err error) {
-	if address == "" {
-		return nil, errors.New("empty address")
-	}
-	logger.Printf("dapr client initializing for: %s", address)
-
 	timeoutSeconds, err := getClientTimeoutSeconds()
 	if err != nil {
 		return nil, err
 	}
 	ctx, cancel := context.WithTimeout(ctx, time.Duration(timeoutSeconds)*time.Second)
+	defer cancel()
+	return NewClientWithAddressContextNoEnvTimeout(ctx, address)
+}
+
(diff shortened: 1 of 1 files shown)


artursouza (changes requested):
We have an approved proposal that let's users configure the timeout via env variable. The name has changed, to become a standard across SDKs. Can this env variable be kept and change the default to be "no timeout" instead of 5s?

artursouza:
Another options is to have the env var override the timeout for the context (if possible).

artursouza:
It is a breaking change. Yaron would need to accept that as a maintainer
here. I am not.

yaron2:
I have knowledge of Go-SDK users that rely on the default 5s timeout for various reasons, so I recommend to not change the default.

client/client.go
```diff
@@ -253,24 +253,31 @@ func NewClientWithAddress(address string) (client Client, err error) {
...
 	}
 	ctx, cancel := context.WithTimeout(ctx, time.Duration(timeoutSeconds)*time.Second)
+	defer cancel()
+	return NewClientWithAddressContextNoEnvTimeout(ctx, address)
+}
+
+// NewClientWithAddressContextNoEnvTimeout instantiates Dapr using specific address (including port).
+// Ignores the DAPR_GRPC_TIMEOUT_SECONDS environment variable and has no
+// default timeout.
+func NewClientWithAddressContextNoEnvTimeout(ctx context.Context, address string) (client Client, err error) {
```
yaron2, line 268: Recommend changing to `NewClientWithAddressContextNoTimeout`

daixiang0:
ping @JoshVanL

client/client.go
JoshVanL: Done.

PR #441, opened 2023-08-24
fix actor serialization

### LINKED ISSUE #317: Serialization error - Actor codec is ignored by client
**Describe the bug**

Custom serialization codecs (with [`codec.SetActorCodec()`](https://github.com/dapr/go-sdk/blob/a38be4e38b7d51406e366f60fc5a526d255f46e1/actor/codec/codec.go#L31)) do not seem to be working. Actor invocations are always encoded as JSON.

Setting and then using a custom codec in a client and serving instances of an application result in the serving application returning a `HTTP 500` error to the Dapr placement server. I suspect it is because the client code always uses JSON to marshal the request: https://github.com/dapr/go-sdk/blob/7c38f5a607d58ee0af93a2f26a061d5c1834203a/client/actor.go#L398

**To Reproduce**

Take the `examples/actor` example and add this code to the `main()` function in the `client` and `serving` packages:

```go
codec.SetActorCodec("custom", func() codec.Codec {
	// Return your custom codec here.
})
```

Add the `config.WithSerializerName("custom")` option to the `ImplActorClientStub()` method call in the `client` package, and to the `RegisterActorImplFactory()` method call in the `serving` package.

Run the applications following the README.

mikeee:
Issue where the serializer isn't used and json is specified to marshal/unmarshal parms

Merging this will close #317

Signed-off-by: mikeee <hey@mike.ee>

### DIFF
--- a/actor/manager/manager.go
+++ b/actor/manager/manager.go
@@ -15,7 +15,6 @@ package manager
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"log"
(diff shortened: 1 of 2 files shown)


yaron2 (approved):

yaron2:
Thanks for this cocntribution @mikeee

PR #457, opened 2023-09-22
add note on unsubscribeConfiguration being deprecated

cicoyle:
Explicitly noting a deprecated function to make it more obvious to end users. Suggestions to wording it differently are welcomed.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -142,8 +142,8 @@ type Client interface {
 	// SubscribeConfigurationItems can subscribe the change of configuration items by storeName and keys, and return subscription id
 	SubscribeConfigurationItems(ctx context.Context, storeName string, keys []string, handler ConfigurationHandleFunction, opts ...ConfigurationOpt) (string, error)
 
-	// This function calls UnsubscribeConfiguration which is deprecated and exists for backwards-compatibility only
-	// The unsubscribe occurs when closing the subscription stream
+	// UnsubscribeConfigurationItems stops the subscription with target store's and ID.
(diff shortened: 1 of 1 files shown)


yaron2 (dismissed):

client/client.go
```diff
@@ -142,6 +142,8 @@ type Client interface {
 	// SubscribeConfigurationItems can subscribe the change of configuration items by storeName and keys, and return subscription id
 	SubscribeConfigurationItems(ctx context.Context, storeName string, keys []string, handler ConfigurationHandleFunction, opts ...ConfigurationOpt) (string, error)
 
+	// This function calls UnsubscribeConfiguration which is deprecated and exists for backwards-compatibility only
+	// The unsubscribe occurs when closing the subscription stream
```
JoshVanL, line 146: Please use `Deprecated: ` prefix in the comment. This is what we do elsewhere and ensures IDEs and linters warn the developer when they are using a deprecated function.

```suggestion
	// UnsubscribeConfigurationItems stops the subscription with target store's and ID.
	// Deprecated: Closing the `SubscribeConfigurationItems` stream (closing the given context) will unsubscribe the client and should be used in favor of `UnsubscribeConfigurationItems`.
```

JoshVanL (changes requested):

client/client.go
cicoyle: Good catch. thx!

daixiang0 (approved):
LGTM

PR #469, opened 2023-11-02
Updates pinned runtime

ItalyPaleAle:
This is needed in components-contrib as we cannot pin an updated runtime in cert tests without an updated Go SDK, due to a complex circular dependency

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ tidy: ## Updates the go modules
 
 .PHONY: test
 test:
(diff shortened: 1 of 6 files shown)


yaron2:
Examples validation is failing

ItalyPaleAle:
> Examples validation is failing

Seems unrelated. Flaky test?

yaron2 (approved):

Issue #379, opened 2023-03-13
Workflow implementation in SDK


RyanLettieri:
**Is your feature request related to a problem? Please describe.**

Related dapr/dapr proposal for workflow: https://github.com/dapr/dapr/issues/4576

**Describe the solution you'd like**
Add GO SDK support for dapr workflow. This will include both the authoring and management functionalities.

ArunNKutty:
Are we planning to have workflows in go as well?

yaron2:
Indeed, that's being planned.

PR #476, opened 2023-11-13
feat: implement workflow management

mikeee:
[WIP]
Implements Workflow management (Alpha + Beta endpoints).

E2E testing is not yet implemented as Authoring is still outstanding.

This does not yet fully close this issue: #379 as authoring is not included in this PR.
Contributes towards the completion of dapr/dapr#7156

- [ ] Testing implemented
- [ ] Checks/Tests passing
- [ ] Documentation updated/improved

### DIFF
--- a/client/workflow.go
+++ b/client/workflow.go
@@ -260,7 +260,7 @@ func (c *GRPCClient) StartWorkflowBeta1(ctx context.Context, req *StartWorkflowR
 
 	var input []byte
 	var err error
-	if (!req.SendRawInput) || (req.Input != nil) {
+	if (!req.SendRawInput) && (req.Input != nil) {
 		input, err = json.Marshal(req.Input)
 		if err != nil {
 			return nil, fmt.Errorf("failed to marshal input: %v", err)
--- a/client/workflow.go
+++ b/client/workflow.go
@@ -70,8 +70,12 @@ func (c *GRPCClient) StartWorkflowAlpha1(ctx context.Context, req *StartWorkflow
 	if req.InstanceID == "" {
 		req.InstanceID = uuid.New().String()
 	}
-	if req.WorkflowComponent == "" || req.WorkflowName == "" {
-		return nil, errors.New("failed to start workflow: WorkflowComponent and WorkflowName must be supplied")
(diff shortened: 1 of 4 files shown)


yaron2:
cc @cgillum

client/workflow.go
```diff
@@ -0,0 +1,310 @@
...
+	EventName         string
+	EventData         []byte // Optional data
+}
+
+// StartWorkflowAlpha1 starts a workflow instance using the alpha1 spec.
+func (c *GRPCClient) StartWorkflowAlpha1(ctx context.Context, req *StartWorkflowRequest) (*StartWorkflowResponse, error) {
+	if req.InstanceID == "" {
+		req.InstanceID = uuid.New().String()
+	}
+	if req.WorkflowComponent == "" || req.WorkflowName == "" {
```
RyanLettieri, line 73: Can we separate this if check into 2 separate checks with separate errors? For a user, if they provide one or the other, they won't know immediately whether or not they didn't supply the WorkflowComponent or the WorkflowName. This comment would apply in all the functions below as well.

yaron2:
Thanks for picking up the review here @RyanLettieri

client/workflow_test.go
RyanLettieri, line 1: One test case I would like to see is passing in an input to the start workflow calls. In other SDKs we've run into an issue where certain input types would lead to an error since they couldn't be parsed correctly. For example, in .NET we needed to serialize the data and convert it into a bytestring. It may not be an issue in GO, but having the test coverage can't hurt.

client/workflow.go
```diff
@@ -0,0 +1,373 @@
...
+type ResumeWorkflowRequest struct {
+	InstanceID        string
+	WorkflowComponent string
+}
+
+type RaiseEventWorkflowRequest struct {
+	InstanceID        string
+	WorkflowComponent string
+	EventName         string
+	EventData         []byte // Optional data
```
RyanLettieri, line 66: Much like the StartWorkflow Request, we most likely need to deal with data serialization. When you get around to testing the passing in of input data to the start workflow requests, please do the same with raise event

mikeee:
Merging these commits into #487

PR #480, opened 2023-12-06
feat: go1.20 and golangci-lint v1.55.2

mikeee:
Brings the package up to go1.20

golangci-lint has been bumped from v1.54.2 to v1.55.2


daixiang0 (approved):
LGTM

PR #481, opened 2023-12-07
fix: update StateConsistency type to match protobufs

mikeee:
Fixes StateConsistency type to match protobufs, at the moment returns undefined for everything due to the logic.

https://github.com/dapr/dapr/blob/ed3417223c5ea4654f38df1fc5abee93e52cea6e/dapr/proto/common/v1/common.proto#L140

### DIFF
--- a/client/state.go
+++ b/client/state.go
@@ -46,6 +46,19 @@ const (
 	StateOperationTypeUpsert OperationType = 1
 	// StateOperationTypeDelete represents delete operation type value.
 	StateOperationTypeDelete OperationType = 2
+
+	// EventualType represents the eventual type value.
+	EventualType = "eventual"
+	// StrongType represents the strong type value.
+	StrongType = "strong"
+	// FirstWriteType represents the first write type value.
+	FirstWriteType = "first-write"
+	// LastWriteType represents the last write type value.
+	LastWriteType = "last-write"
+	// UpsertType represents the upsert type value.
+	UpsertType = "upsert"
+	// DeleteType represents the delete type value.
+	DeleteType = "delete"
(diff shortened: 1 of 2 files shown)


client/state_test.go
```diff
@@ -29,12 +29,24 @@ const (
...
-	assert.Equal(t, UndefinedType, op.String())
-	var c StateConcurrency = -1
-	assert.Equal(t, UndefinedType, c.String())
-	var d StateConsistency = -1
-	assert.Equal(t, UndefinedType, d.String())
+	t.Run("test operation types", func(t *testing.T) {
+		var a OperationType = -1
+		assert.Equal(t, UndefinedType, a.String())
+		a = 2
+		assert.Equal(t, "delete", a.String())
```
daixiang0, line 36: Use constant like `UndefinedType`.

client/state.go
```diff
@@ -46,6 +46,19 @@ const (
...
+	// EventualType represents the eventual type value.
+	EventualType = "eventual"
+	// StrongType represents the strong type value.
+	StrongType = "strong"
+	// FirstWriteType represents the first write type value.
+	FirstWriteType = "first-write"
+	// LastWriteType represents the last write type value.
+	LastWriteType = "last-write"
+	// UpsertType represents the upsert type value.
+	UpsertType = "upsert"
```
daixiang0, line 59: Is it reserved and not used yet?

mikeee: Added testing for it.

mikeee:
Looks like I accidentally committed this to #480 and it was merged. With the latest it implements the recommendations and further unit tests.

yaron2 (approved):

PR #483, opened 2023-12-14
perf: enhanced configuration example

sadath-12:
Eliminated ```time.sleep``` with ```sync.WaitGroup``` so ```SubscribeConfigurationItems``` can take proper time to finish and also updated to more appropriate logs

### DIFF
--- a/examples/configuration/main.go
+++ b/examples/configuration/main.go
@@ -20,7 +20,7 @@ func addItems(wg *sync.WaitGroup) {
 	// set config value
 	client.Set(context.Background(), "mykey", "myConfigValue", -1)
 	ticker := time.NewTicker(time.Second)
-	wg.Add(3*5)
+	wg.Add(3 * 5)
 	go func() {
 		for i := 0; i < 5; i++ {
 			<-ticker.C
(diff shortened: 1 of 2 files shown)


examples/configuration/main.go
```diff
@@ -4,21 +4,23 @@ import (
...
-func init() {
+func addItems(wg *sync.WaitGroup) {
 	opts := &redis.Options{
 		Addr: "127.0.0.1:6379",
 	}
 	client := redis.NewClient(opts)
 	// set config value
 	client.Set(context.Background(), "mykey", "myConfigValue", -1)
 	ticker := time.NewTicker(time.Second)
+	wg.Add(3*5)
```
daixiang0, line 23: hardcode?

sadath-12: Nope 😅 Since the loop runs 5 times and for every run there is 3 entry its 5*3 to make sure all entries that's send get recieved

mikeee (changes requested):
I agree with this moving towards no time dependency where possible, but it could also be argued that this is a form of validation to ensure that there are no excessive waits.

The mechanical markdown validation hasn't been updated for your revision.

I'm not too sure on the verbiage change being any better than previously. I'd like to see the "with" prefix removed back to the original state

sadath-12:
> I agree with this moving towards no time dependency where possible, but it could also be argued that this is a form of validation to ensure that there are no excessive waits.
> 
> The mechanical markdown validation hasn't been updated for your revision.
> 
> I'm not too sure on the verbiage change being any better than previously. I'd like to see the "with" prefix removed back to the original state

Hmm yeah , also making sure whatever we are passing we are getting it back correctly whereas before the information retrieved was missing from what's sent

daixiang0 (approved):
LGTM

Issue #449, opened 2023-09-14
build: add PR template


sicoyle:
The Dapr Go SDK repo should have a PR template for when end users create PRs contributing to the project.

**Describe the solution you'd like**
We should abide by a standardized PR template across all SDK repositories.

To implement, we should add a `pull_request_template.md` file. Something similar to what we have in `dapr/dapr` (https://github.com/dapr/dapr/blob/master/.github/pull_request_template.md), but also considering if we need special additions for SDK specific repositories.

**Additional context**
This will help to improve contributors to easily communicate their changes, as well as maintainers to easily understand what the PR is doing.

sicoyle:
Java SDK PR template to use as inspiration as well: https://github.com/dapr/java-sdk/blob/master/.github/pull_request_template.md

sadath-12:
Can be closed

mikeee:
Closed with #484

PR #484, opened 2023-12-14
PR template added

sadath-12:
Added pr template as suggested in #449

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,17 @@
+# Description
+
+_Please explain the changes you've made_
+
+## Issue reference
+
+We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.
+
+Please reference the issue this PR will close: #_[issue number]_
+
+## Checklist
+
+Please make sure you've completed the relevant tasks for this PR, out of the following list:
+
+* [ ] Code compiles correctly
+* [ ] Created/updated tests
+* [ ] Extended the documentation
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,6 @@
 # Description
 
-_Please explain the changes you've made_
-
 ## Issue reference
-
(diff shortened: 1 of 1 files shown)


daixiang0 (approved):

mikeee (changes requested):
I believe the descriptions or context should be in a comment so can be ignored on open rather than having to remove/delete the placeholder text.

sadath-12:
> Me

> I believe the descriptions or context should be in a comment so can be ignored on open rather than having to remove/delete the placeholder text.

Just felt if we need to make sure the contributor must always provide information of what he did instead of making it optional . Please let me know if I am thinking it in right way

.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
+# Description
+
+_Please explain the changes you've made_
```
daixiang0, line 3: Hide it.

.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Issue reference
+
+We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.
```
daixiang0, line 7: Same as above.

sadath-12:
thanks done @daixiang0

mikeee:
To clarify my opinion is that we should comment out the placeholder/descriptive text so that we provide an indication of what to include and does not include it.

mikeee:
To clarify - my opinion is that we should comment out the placeholder/descriptive text so that we provide an indication of what to include and does not include it.

daixiang0:
> To clarify - my opinion is that we should comment out the placeholder/descriptive text so that we provide an indication of what to include and does not include it.

Agree, @sadath-12 please comment rather than delete.

sadath-12:
yeah I get it now thanks @daixiang0  @mikeee

daixiang0 (approved):

PR #487, opened 2023-12-17
feat: workflow authoring and management support

### LINKED ISSUE #379: Workflow implementation in SDK

mikeee:
Implements Workflow authoring using `durabletask-go` and also an interface to the dapr grpc management endpoints.

Contributes towards the completion of dapr/dapr#7156 and closes #379

This PR also exposes the GRPC connection from the dapr client.

Requires documentation in https://github.com/dapr/docs/pull/3895

- [x] Testing implemented (E2E)
- [x] Checks/Tests passing
- [x] Documentation updated/improved


client/client.go
```diff
@@ -209,8 +209,52 @@ type Client interface {
...
+	TerminateWorkflowAlpha1(ctx context.Context, req *TerminateWorkflowRequest) error
+
+	// PauseWorkflowAlpha1 pauses a workflow.
+	PauseWorkflowAlpha1(ctx context.Context, req *PauseWorkflowRequest) error
+
+	// ResumeWorkflowAlpha1 resumes a workflow.
+	ResumeWorkflowAlpha1(ctx context.Context, req *ResumeWorkflowRequest) error
+
+	// RaiseEventWorkflowAlpha1 raises an event for a workflow.
+	RaiseEventWorkflowAlpha1(ctx context.Context, req *RaiseEventWorkflowRequest) error
```
mukundansundar, line 231: can remove alpha endpoints.

examples/workflow/README.md
```diff
@@ -0,0 +1,66 @@
...
+         --log-level debug \
+         --resources-path ./config \
+         -- go run ./main.go
+```
+
+<!-- END_STEP -->
+
+## Result
+
+- workflow
```
mukundansundar, line 47: can be removed.

mukundansundar (commented):
- consider adding authoring SDK
- should generics support be added instead of `any interface{}` as if there are plans to add it later on that might be a breaking change?

cc @yaron2 wdyt?

examples/workflow/README.md
mukundansundar: Please do not remove the who result output ... only this line workflow can be removed.

mikeee: Reverted the change, thanks for catching.

yaron2:
Amazing! Thanks @mikeee for adding the authoring code.

yaron2:
@cgillum for review

mikeee:
Apologies for the excess formatting - need to tone down the nvim plugins.

workflow/client.go
```diff
@@ -0,0 +1,158 @@
...
+		return "", errors.New("no workflow specified")
+	}
+	workflowID, err := c.taskHubClient.ScheduleNewOrchestration(ctx, workflow, opts...)
+	if err != nil {
+		return "", err
+	}
+	return string(workflowID), nil
+}
+
+func (c *client) FetchWorkflowMetadata(ctx context.Context, id string, opts ...api.FetchOrchestrationMetadataOptions) (*api.OrchestrationMetadata, error) {
```
cgillum, line 95: I noticed that you're exposing the `durabletask-go` types directly in these APIs rather than wrapping them with Dapr-specific types. I don't necessarily have an issue with this, but just wanted to confirm whether this was intentional since it will couple the two abstractions more tightly.

workflow/client.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return c.taskHubClient.RaiseEvent(ctx, api.InstanceID(id), eventName, opts...)
+}
+
+func (c *client) SuspendWorkflow(ctx context.Context, id, reason string) error {
+	if id == "" {
+		return errors.New("no workflow id specified")
+	}
+	if reason == "" {
+		return errors.New("no reason specified")
```
cgillum, line 138: Specifying a reason should be optional.

workflow/context.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/microsoft/durabletask-go/task"
+)
+
+type Context struct {
```
cgillum, line 11: Should this be named `WorkflowContext` to avoid collisions or confusion with `context.Context`?

workflow/runtime.go
```diff
@@ -0,0 +1,125 @@
...
+		aCtx := ActivityContext{ctx: ctx}
+
+		return a(aCtx)
+	}
+}
+
+func (wr *WorkflowRuntime) RegisterActivity(a Activity) error {
+	wrappedActivity := wrapActivity(a)
+
+	// get decorator for activity
```
cgillum, line 95: What do you mean by "decorator" here? Aren't we just getting the name from the name of the function?

workflow/runtime.go
```diff
@@ -0,0 +1,125 @@
...
+
+	err = wr.tasks.AddActivityN(name, wrappedActivity)
+	return err
+}
+
+func (wr *WorkflowRuntime) Start() error {
+	// go func start
+	go func() {
+		defer wr.close()
+		err := wr.client.StartWorkItemListener(context.Background(), wr.tasks)
```
cgillum, line 109: I wonder if it's better to do this on the calling goroutine so that we can return the error back to the caller rather than doing a `log.Fatalf`.

examples/workflow/main.go
```diff
@@ -0,0 +1,328 @@
...
+	fmt.Println("TestActivity registered")
+
+	var wg sync.WaitGroup
+
+	// Start workflow runner
+	fmt.Println("runner started")
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := wr.Start(); err != nil {
```
cgillum, line 45: The internal implementation of `Start()` already creates a background goroutine, so I don't think we need to do this call to `Start()` in a goroutine. Instead, just call it directly.

examples/workflow/main.go
```diff
@@ -0,0 +1,328 @@
...
+			log.Fatal(err)
+		}
+	}()
+
+	time.Sleep(time.Second * 5)
```
cgillum, line 50: Is this sleep necessary?

workflow/runtime.go
```diff
@@ -0,0 +1,125 @@
...
+		if err != nil {
+			log.Fatalf("failed to start work stream: %v", err)
+		}
+		log.Println("work item listener started")
+		<-wr.quit
+		log.Println("work item listener shutdown")
```
cgillum, line 115: In order to fully complete the shutdown, we need to also cancel the `context.Context` object that was passed to `StartWorkItemListener`. You currently are using `context.Background`, which unfortunately won't work (you'll need to create a new cancellable context).

workflow/state.go
```diff
@@ -0,0 +1,43 @@
...
+		"PENDING",
+		"SUSPENDED",
+	}
+	if s > StatusSuspended || s < StatusRunning {
+		return "UNKNOWN"
+	}
+	return status[s]
+}
+
+type WorkflowState struct {
```
cgillum, line 36: Are we using these structs anywhere? It seems they're only referenced here and in `state_test.go`.

workflow/workflow.go
```diff
@@ -0,0 +1,22 @@
...
+	RuntimeStatus          Status          `json:"status"`
+	CreatedAt              time.Time       `json:"createdAt"`
+	LastUpdatedAt          time.Time       `json:"lastUpdatedAt"`
+	SerializedInput        string          `json:"serializedInput"`
+	SerializedOutput       string          `json:"serializedOutput"`
+	SerializedCustomStatus string          `json:"serializedCustomStatus"`
+	FailureDetails         *FailureDetails `json:"failureDetails"`
+}
+
+type FailureDetails struct {
```
cgillum, line 17: Seems like these structs also aren't being used?

go.mod
```diff
@@ -8,17 +8,23 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/golang/protobuf v1.5.3
 	github.com/google/uuid v1.3.1
+	github.com/microsoft/durabletask-go v0.3.1
```
cgillum, line 11: Consider bumping this up to `v0.4.0`

cgillum (commented):
Happy to see this PR! Here is my initial batch of comments.

workflow/client.go
```diff
@@ -0,0 +1,158 @@
...
+
+type client struct {
+	taskHubClient *durabletaskclient.TaskHubGrpcClient
+}
+
+func WithInstanceID(id string) api.NewOrchestrationOptions {
+	return api.WithInstanceID(api.InstanceID(id))
+}
+
+// TODO: Implement WithOrchestrationIdReusePolicy
```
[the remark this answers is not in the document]
mikeee: https://github.com/dapr/go-sdk/issues/498

Issue opened

workflow/client.go
```diff
@@ -0,0 +1,158 @@
...
+
+// TODO: Implement mocks
+
+func NewClient() (client, error) { // TODO: Implement custom connection
+	daprClient, err := dapr.NewClient()
+	if err != nil {
+		return client{}, err
+	}
+
+	taskHubClient := durabletaskclient.NewTaskHubGrpcClient(daprClient.GrpcClientConn(), backend.DefaultLogger())
```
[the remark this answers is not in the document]
mikeee: Issue opened: https://github.com/dapr/go-sdk/issues/499

workflow/runtime.go
mikeee: Yep, this was the previous name for the function.

client/workflow.go
```diff
@@ -0,0 +1,245 @@
...
+	SendRawData       bool // Set to True in order to disable serialization on the data
+}
+
+// StartWorkflowBeta1 starts a workflow using the beta1 spec.
+func (c *GRPCClient) StartWorkflowBeta1(ctx context.Context, req *StartWorkflowRequest) (*StartWorkflowResponse, error) {
+	if req.InstanceID == "" {
+		req.InstanceID = uuid.New().String()
+	}
+	if req.WorkflowComponent == "" {
+		return nil, errors.New("failed to start workflow: WorkflowComponent must be supplied")
```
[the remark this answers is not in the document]
mikeee: Agreed it should in fact be optional, I've added a fallback to `dapr`

workflow/client.go
mikeee: Indeed I was thinking about this, I have no specific leaning either way but in this instance it made more sense in my head to expose these types rather than wrapping them for the workflow client. I'm happy to wrap everything if this is a blocker.

workflow/worker.go
```diff
@@ -0,0 +1,148 @@
...
+	quit   chan bool
+	close  func()
+	cancel context.CancelFunc
+}
+
+type Workflow func(ctx *WorkflowContext) (any, error)
+
+type Activity func(ctx ActivityContext) (any, error)
+
+func NewWorker() (*WorkflowWorker, error) {
```
famarting, line 48: can you make it so a worker can be initialized using an existing dapr client? I just noticed looking at the example included in this PR that it would initialize 3 different dapr clients, one for the worker, another for just using it and another for the workflow client... and I was wondering if all of that could just be reduced to a single dapr client

mikeee: Added an option so that a dapr client can be passed to create the worker.

famarting: thank you, would it make sense to do the same to the same in `workflow/client.go` ? and btw, just out of curiosity, why `workflow/client.go` doesn't it serve the same purpose as the management functions added here `client/client.go`?

famarting (approved):

famarting:
do you know if its possible to add support for setting custom statuses? like here https://github.com/dapr/dotnet-sdk/blob/master/examples/Workflow/WorkflowConsoleApp/Workflows/OrderProcessingWorkflow.cs#L51

mikeee:
> do you know if its possible to add support for setting custom statuses? like here https://github.com/dapr/dotnet-sdk/blob/master/examples/Workflow/WorkflowConsoleApp/Workflows/OrderProcessingWorkflow.cs#L51

Not sure about the setting of custom statuses, I don't believe so at this stage - but I'll have a look into it.

client/workflow.go
```diff
@@ -0,0 +1,272 @@
...
+		req.WorkflowComponent = DefaultWorkflowComponent
+	}
+	if req.WorkflowName == "" {
+		return nil, errors.New("failed to start workflow: WorkflowName must be supplied")
+	}
+
+	var input []byte
+	var err error
+	if req.SendRawInput {
+		input = req.Input.([]byte)
```
cgillum, line 104: Should we check for a casting error and return a friendly error accordingly?

client/workflow.go
```diff
@@ -0,0 +1,272 @@
...
+		WorkflowComponent: req.WorkflowComponent,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get workflow status: %v", err)
+	}
+	if resp.GetCreatedAt() == nil {
+		resp.CreatedAt = timestamppb.Now()
+	}
+	if resp.GetLastUpdatedAt() == nil {
+		resp.LastUpdatedAt = timestamppb.Now()
```
cgillum, line 146: I wonder if it makes more sense to leave these timestamp values as nil or as some zero value rather than setting them to `timestamppb.Now()`, which could be confusing.

client/workflow.go
```diff
@@ -0,0 +1,272 @@
...
+	if req.WorkflowComponent == "" {
+		req.WorkflowComponent = DefaultWorkflowComponent
+	}
+	if req.EventName == "" {
+		return errors.New("failed to raise event on workflow: EventName must be supplied")
+	}
+	var eventData []byte
+	var err error
+	if req.SendRawData {
+		eventData = req.EventData.([]byte)
```
cgillum, line 244: Same question about whether we should handle a failure here gracefully.

client/workflow.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	return nil
+}
+
+func marshalInput(input any) (data []byte, err error) {
+	if input == nil {
+		return nil, nil
+	}
+	if _, typeByteArray := input.([]byte); typeByteArray {
+		return input.([]byte), nil
```
cgillum, line 269: This feels a bit problematic. We already have `SendRawData` property which is supposed to handle this case, making this redundant. But also, it prevents workflow authors from accepting a JSON-serialized byte array. It's a corner case, but represents a behavior that's different from other SDKs and would be hard to debug. I recommend removing this check and always calling `json.Marshal` for non-nil, non-raw inputs, including byte arrays.

workflow/activity_context.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil
+	}
+}
+
+func marshalData(input any) ([]byte, error) {
+	if input == nil {
+		return nil, nil
+	}
+	if _, typeByteArray := input.([]byte); typeByteArray {
+		return input.([]byte), nil
```
cgillum, line 67: Same comment - let's not treat `[]byte` special here. Users should instead use `ActivityRawInput` if they want to send raw payloads.

workflow/context.go
```diff
@@ -0,0 +1,94 @@
...
+	return wfc.orchestrationContext.CreateTimer(duration)
+}
+
+func (wfc *WorkflowContext) WaitForExternalEvent(eventName string, timeout time.Duration) task.Task {
+	if eventName == "" {
+		return nil
+	}
+	if timeout == 0 {
+		// default to 10 seconds
+		timeout = time.Second * 10
```
cgillum, line 84: The base SDK already handles the `0` case, so I suggest not trying to handle it here unless you have a strong reason to. You can see the implementation [here](https://github.com/microsoft/durabletask-go/blob/fb5c4c05729d85feb9d4b4bbe7246a8f14c9688d/task/orchestrator.go#L310-L313), which treats zero as zero rather than assigning an arbitrary default value.

workflow/worker.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	err = ww.tasks.AddActivityN(name, wrappedActivity)
+	return err
+}
+
+func (ww *WorkflowWorker) Start() error {
+	// go func start
+	errChan := make(chan error)
+	go func() {
```
cgillum, line 149: Can this be simplified by removing the goroutine? It doesn't seem to be needed.

workflow/worker.go
```diff
@@ -0,0 +1,174 @@
...
+	name, err := getFunctionName(a)
+	if err != nil {
+		return fmt.Errorf("failed to get activity decorator: %v", err)
+	}
+
+	err = ww.tasks.AddActivityN(name, wrappedActivity)
+	return err
+}
+
+func (ww *WorkflowWorker) Start() error {
```
cgillum, line 146: I suggest adding some comments to the public methods to help users understand what they are for and what the behavior's are. For example, it would be good to mention that `Start()` is non-blocking.

workflow/worker.go
```diff
@@ -0,0 +1,174 @@
...
+	// go func start
+	errChan := make(chan error)
+	go func() {
+		defer ww.close()
+		ctx, cancel := context.WithCancel(context.Background())
+		err := ww.client.StartWorkItemListener(ctx, ww.tasks)
+		if err != nil {
+			cancel()
```
cgillum, line 154: If I'm reading this correctly, the use of `cancel()` here doesn't actually do anything because the call that you passed the corresponding `ctx` with already failed with an error, thus shouldn't need to be canceled.

workflow/worker.go
```diff
@@ -0,0 +1,174 @@
...
+		log.Println("work item listener shutdown")
+	}()
+	err := <-errChan
+	return err
+}
+
+func (ww *WorkflowWorker) Shutdown() error {
+	ww.cancel()
+	// send close signal
+	ww.quit <- true
```
cgillum, line 171: If I'm reading the code correctly, it seems the only thing that invoking `ww.quit` does is cause the message `"work item listener shutdown` to get logged. Rather than using a channel for that, why not just log it here?

cgillum (commented):
This is looking much better. A few more things I found in this latest review iteration.

client/workflow.go
mikeee: I'll leave them as nil

client/workflow.go
mikeee: Whoops, missed this. I'll handle errors gracefully.

yaron2:
Looks like we can merge this. @cgillum

client/actor.go
```diff
@@ -66,7 +66,7 @@ func (c *GRPCClient) InvokeActor(ctx context.Context, in *InvokeActorRequest) (o
 		Data:      in.Data,
 	}
 
-	resp, err := c.protoClient.InvokeActor(c.withAuthToken(ctx), req)
+	resp, err := c.protoClient.InvokeActor(ctx, req)
```
cgillum, line 69: Not sure why I'm seeing these unrelated changes in this PR?

client/workflow.go
```diff
@@ -0,0 +1,271 @@
...
+		req.WorkflowComponent = DefaultWorkflowComponent
+	}
+	if req.EventName == "" {
+		return errors.New("failed to raise event on workflow: EventName must be supplied")
+	}
+	var eventData []byte
+	var err error
+	if req.SendRawData {
+		if reflect.ValueOf(req.EventData).Type() != typeBytes {
+			return errors.New("failed to raise event on workflow: sendrawinput is true however, eventData is not a byte slice")
```
cgillum, line 244:
```suggestion
			return errors.New("failed to raise event on workflow: SendRawData is true however, eventData is not a byte slice")
```

client/workflow.go
```diff
@@ -0,0 +1,271 @@
...
+	if req.WorkflowComponent == "" {
+		req.WorkflowComponent = DefaultWorkflowComponent
+	}
+	if req.WorkflowName == "" {
+		return nil, errors.New("failed to start workflow: WorkflowName must be supplied")
+	}
+
+	var input []byte
+	var err error
+	if req.SendRawInput {
```
cgillum, line 105: The other API calls this `SendRawData`. Should we make these consistent?

cgillum (commented):
Working my way through this. Adding a few comments now before I finish reviewing.

client/workflow.go
cgillum: On second thought, making these consistent probably isn't important. One is for workflow *input*, the other is for event *data*.

client/workflow.go
```diff
@@ -0,0 +1,271 @@
...
+	if req.WorkflowComponent == "" {
+		req.WorkflowComponent = DefaultWorkflowComponent
+	}
+	if req.EventName == "" {
+		return errors.New("failed to raise event on workflow: EventName must be supplied")
+	}
+	var eventData []byte
+	var err error
+	if req.SendRawData {
+		if reflect.ValueOf(req.EventData).Type() != typeBytes {
```
cgillum, line 243: This probably works too, but I thought it was more conventional in Go to do something like this:
```go
if eventData, ok = req.EventData.([]byte); !ok {
    return errors.New("...")
}
```

workflow/context.go
```diff
@@ -0,0 +1,102 @@
...
+	return wfc.orchestrationContext.CurrentTimeUtc
+}
+
+// IsReplaying returns whether the workflow is replaying.
+func (wfc *WorkflowContext) IsReplaying() bool {
+	return wfc.orchestrationContext.IsReplaying
+}
+
+// CallActivity returns a completable task for a given activity.
+// NOTE: The task must be invoked with the .Await(output interface{}) method.
```
cgillum, line 54: Regarding these NOTEs, users can choose to not await the tasks at call time. This is most common when doing the fan-out/fan-in pattern, where you want multiple tasks to run in parallel. I wonder if we should change these comments.

cgillum (commented):
Just a couple super-minor things. The most important is probably changing the wording in the documentation comments regarding `.Await(...)` as I'm a bit worried that users might misunderstand how tasks can be used.

client/actor.go
mikeee: I had to pull in main and fix conflicts from #502. This commit history is messy as a result of this conflict resolution but it does not differ from main.

client/workflow.go
mikeee: Makes sense - implemented.

workflow/context.go
mikeee: I think the note update improves clarity slightly

client/workflow.go
```diff
@@ -240,10 +240,10 @@ func (c *GRPCClient) RaiseEventWorkflowBeta1(ctx context.Context, req *RaiseEven
 	var eventData []byte
 	var err error
 	if req.SendRawData {
-		if reflect.ValueOf(req.EventData).Type() != typeBytes {
-			return errors.New("failed to raise event on workflow: sendrawinput is true however, eventData is not a byte slice")
+		var ok bool
+		if eventData, ok = req.EventData.([]byte); !ok {
```
cgillum, line 244: For consistency, we should make the same change in `StartWorkflowBeta1`.

workflow/context.go
```diff
@@ -64,7 +65,8 @@ func (wfc *WorkflowContext) CallActivity(activity interface{}, opts ...callActiv
 }
 
 // CallChildWorkflow returns a completable task for a given workflow.
-// NOTE: The task must be invoked with the .Await(output interface{}) method.
+// NOTE: The task must be invoked with the .Await(output interface{}) method, this does not need to
+// be done at call time for example when using the fan-out/fan-in pattern.
```
cgillum, line 69: I still don't feel great about this explanation. Here's my suggestion. I apologize that it's a bit verbose.
```go
// You must call Await(output any) on the returned Task to block the workflow and wait for the task to complete.
// The value passed to the Await method must be a pointer or can be nil to ignore the returned value.
// Alternatively, tasks can be awaited using the task.WhenAll or task.WhenAny methods, allowing the workflow
// to block and wait for multiple tasks at the same time.
```

cgillum (approved):
I'm signing off on this but with a couple final recommendations.

I appreciate the hard work that went into this. :)

workflow/context.go
mikeee: Make no apologies for verbosity, thank you for clarifying!

mikeee:
@yaron2 for your eyes

yaron2:
@mikeee please resolve the conflicts and we can merge

mikeee:
@yaron2 good to go, not sure why codecov is unhappy but 🤷

yaron2 (approved):

PR #491, opened 2024-01-03
chore: replaced with dapr-kit logger

sadath-12:
# Description

Currently we use ```log``` package for logging whereas we have our own logger available at ```github.com/dapr/kit/logger``` which we should make use it

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


.github/workflows/dapr-bot/main.go
```diff
@@ -2,12 +2,15 @@ package main
```
daixiang0, line 2: Notice that it is the bot, it should not depend on kit.

sadath-12: oh yeah , thanks for that

daprdocs/content/en/go-sdk-docs/go-client/_index.md
mikeee, line 1: I don't think the examples should have this change made, the standard library does not include these methods and there are no explicit imports specified.

mikeee, line 1: I don't think the documentation should have this change made, the standard library does not include these methods and there are no explicit imports specified.

client/pubsub.go
```diff
@@ -102,7 +101,7 @@ func PublishEventWithRawPayload() PublishEventOption {
 // PublishEventfromCustomContent serializes an struct and publishes its contents as data (JSON) onto topic in specific pubsub component.
 // Deprecated: This method is deprecated and will be removed in a future version of the SDK. Please use `PublishEvent` instead.
 func (c *GRPCClient) PublishEventfromCustomContent(ctx context.Context, pubsubName, topicName string, data interface{}) error {
-	log.Println("DEPRECATED: client.PublishEventfromCustomContent is deprecated and will be removed in a future version of the SDK. Please use `PublishEvent` instead.")
+	log.Infof("DEPRECATED: client.PublishEventfromCustomContent is deprecated and will be removed in a future version of the SDK. Please use `PublishEvent` instead.")
```
mikeee, line 104: Unused formatter

actor/manager/manager.go
```diff
@@ -251,7 +250,7 @@ func suitableMethods(typ reflect.Type) map[string]*MethodType {
 	for m := 0; m < typ.NumMethod(); m++ {
 		method := typ.Method(m)
 		if mt, err := suiteMethod(method); err != nil {
-			log.Printf("method %s is illegal, err = %s, just skip it", method.Name, err)
+			log.Infof("method %s is illegal, err = %s, just skip it", method.Name, err)
```
mikeee, line 253: `err` being formatted as a string rather than value (although a nil value should in theory never reach this point).

examples/hello-world/README.md
mikeee, line 1: The examples should remain mostly the same - can you modify to update the substring matches rather than removing them to pass validations?

sadath-12: ya I thought about . but It is unnecessary right ? . I mean we should just care about the strings that are specific to those examples rather than replicating server starting and many other things everywhere ?

mikeee (changes requested):
I would like to see the logger name kept, we shouldn't have unnecessary package name collisions especially with the standard library.

The service grpc and http documentation should use the standard library / be open to the user.

sadath-12:
> I would like to see the logger name kept, we shouldn't have unnecessary package name collisions especially with the standard library.
> 
> The service grpc and http documentation should use the standard library / be open to the user.

you mean to change the name of  log to something else```var log = logger.NewLogger("dapr.actor.manager")```

mikeee:
> > I would like to see the logger name kept, we shouldn't have unnecessary package name collisions especially with the standard library.
> > The service grpc and http documentation should use the standard library / be open to the user.
> 
> you mean to change the name of log to something else`var log = logger.NewLogger("dapr.actor.manager")`

Yep, I just think it looks cleaner than using a naming convention that clashes with the standard library where it doesn't interface cleanly.

sadath-12:
@mikeee can we think in this way .. it had same name so I was able to quickly migrate to new liberary . so maybe that's a plus point over using different names . just thinkering for this let me know if you agree or else ill change

sadath-12:
ok so shall we do like for example ```logClient``` in the client package

mikeee:
I'm not entirely sure about using `kit/logger` in a user written application would be preferable. 
@qustavo what are your thoughts about this change?

The recent change made by @qustavo allows for a user-specified logger as part of the client created, perhaps this addresses the need to migrate entirely to `kit/logger` and not change the expected output/functionality of the current logging implementation

qustavo:
> I'm not entirely sure about using `kit/logger` in a user written application would be preferable. @qustavo what are your thoughts about this change?

From what I can see, the logging requirements for the sdk are minimal. We `log.Print[f,ln]` and `log.Fatal[f]` and the current logging implementation is sufficient. That being said, we could support a custom logger injection if `SetLogger` accepted an interface rather than a concrete implementation, so that you can bring your own logger, although that might be overkill.

In general I'd wait until the need for a more elaborated logging strategy is needed.

@sadath-12 is there a use case for using `kit/logger` that I'm overlooking?

https://github.com/dapr/go-sdk/blob/7e5ee73aa842660f3bb2c958947d1faa3f665022/client/client.go#L68

PR #496, opened 2024-01-16
allow to customize the default logger

qustavo:
# Description

The following allows the user of the SDK to customize the global logger, as it might break some application logging format. This way one can streamline all the program logging output.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -61,6 +61,18 @@ var (
 	defaultClient Client
 )
 
+// SetLogger sets the global logger for the Dapr client.
+// The default logger has a destination of os.Stdout, SetLogger allows you to
+// optionally specify a custom logger (with a custom destination).
+// To disable client logging entirely, use a nil argument e.g.: client.SetLogger(nil)
+func SetLogger(l *log.Logger) {
+	if l == nil {
+		l = log.New(io.Discard, "", 0)
+	}
+
+	logger = l
+}
+
 // Client is the interface for Dapr client implementation.
 //
 //nolint:interfacebloat


client/client.go
```diff
@@ -61,6 +61,17 @@ var (
 	defaultClient Client
 )
 
+// SetLogger sets the global logger for the Dapr client.
+// By default log.New(os.Stdout, "", 0) is used.
+// To disable logging, set the logger to nil.
+func SetLogger(l *log.Logger) {
+	if l == nil {
+		l = log.New(io.Discard, "", 0)
```
daixiang0, line 69: It is not what the comment says.

qustavo: what do you mean?

mikeee: I think the comment may come across as slightly confusing

client/client.go
```diff
@@ -61,6 +61,17 @@ var (
 	defaultClient Client
 )
 
+// SetLogger sets the global logger for the Dapr client.
+// By default log.New(os.Stdout, "", 0) is used.
+// To disable logging, set the logger to nil.
```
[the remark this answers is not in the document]
qustavo: much better indeed

client/client.go
qustavo: added your suggestion.

mikeee (approved):
This is a non-breaking minor change that extends the usability of the go-sdk. Looks good to me.

mikeee (approved):
lgtm

qustavo:
@mikeee @daixiang0 : how can we get this merged?

yaron2 (approved):

yaron2:
Thanks for your contribution @qustavo

PR #502, opened 2024-01-29
refactor setting of api token

famarting:
# Description

This PR refactors the way the API Token is handled, now instead of having to manually set to token in the context by using `withAuthToken` there are two interceptors that will set the token automatically.

This new approach will guarantee that every new function added to the client will automatically make use of the api token if its set, unlike the current approach, where actually there were a few functions that were not using `withAuthToken` and therefore wouldn't work properly if the api token was set...

Another example of the benefits and the need of this refactor is the use of the grpc connection that is being made in this PR  https://github.com/dapr/go-sdk/pull/487 , since for workflows the grpc connection is used directly and we won't control the calls made by that durabletask grpc client we need to transparently set the api token i.e with the interceptors I added

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -286,9 +286,13 @@ func NewClientWithAddressContext(ctx context.Context, address string) (client Cl
 		return nil, fmt.Errorf("error parsing address '%s': %w", address, err)
 	}
 
+	at := &authToken{}
+
 	opts := []grpc.DialOption{
 		grpc.WithUserAgent(userAgent()),
 		grpc.WithBlock(),
+		authTokenUnaryIterceptor(at),
+		authTokenStreamIterceptor(at),
 	}
 
 	if parsedAddress.TLS {
@@ -307,11 +311,8 @@ func NewClientWithAddressContext(ctx context.Context, address string) (client Cl
 	if err != nil {
 		return nil, fmt.Errorf("error creating connection to '%s': %w", address, err)
 	}
-	if hasToken := os.Getenv(apiTokenEnvVarName); hasToken != "" {
-		logger.Println("client uses API token")
-	}
 
-	return NewClientWithConnection(conn), nil
+	return NewClientWithConnection(conn, at), nil
 }
 
 func getClientTimeoutSeconds() (int, error) {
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -334,36 +335,78 @@ func NewClientWithSocket(socket string) (client Client, err error) {
...
-	if hasToken := os.Getenv(apiTokenEnvVarName); hasToken != "" {
-		logger.Println("client uses API token")
-	}
-	return NewClientWithConnection(conn), nil
+	return NewClientWithConnection(conn, at), nil
 }
 
 // NewClientWithConnection instantiates Dapr client using specific connection.
-func NewClientWithConnection(conn *grpc.ClientConn) Client {
+func NewClientWithConnection(conn *grpc.ClientConn, authToken *authToken) Client {
```
famarting, line 355: this is a breaking change, I'll fix it

client/client.go
```diff
@@ -334,36 +335,82 @@ func NewClientWithSocket(socket string) (client Client, err error) {
...
-	if hasToken := os.Getenv(apiTokenEnvVarName); hasToken != "" {
-		logger.Println("client uses API token")
-	}
-	return NewClientWithConnection(conn), nil
+	return newClientWithConnection(conn, at), nil
 }
 
-// NewClientWithConnection instantiates Dapr client using specific connection.
-func NewClientWithConnection(conn *grpc.ClientConn) Client {
+func newClientWithConnection(conn *grpc.ClientConn, authToken *authToken) Client {
```
mikeee, line 354:
```suggestion
func NewClientWithConnection(conn *grpc.ClientConn, authToken *authToken) Client {
```

client/client.go
```diff
@@ -334,36 +335,82 @@ func NewClientWithSocket(socket string) (client Client, err error) {
...
+	return grpc.WithUnaryInterceptor(func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		token := authToken.get()
+		if token != "" {
+			ctx = metadata.NewOutgoingContext(ctx, metadata.Pairs(apiTokenKey, token))
+		}
+		return invoker(ctx, method, req, reply, cc, opts...)
+	})
+}
+
+func authTokenStreamIterceptor(authToken *authToken) grpc.DialOption {
```
sicoyle, line 399:
```suggestion
func authTokenStreamInterceptor(authToken *authToken) grpc.DialOption {
```

client/client.go
yaron2, line 354: This is a breaking change

mikeee (commented):
Should I be able to explicitly pass a custom token?

famarting:
> Should I be able to explicitly pass a custom token?

that was already possible, and still will be, with the function `WithAuthToken`

yaron2 (approved):

PR #503, opened 2024-02-07
Update github.com/dapr/dapr to v1.13.0-rc.1

JoshVanL:


yaron2:
This is incompatible with Go 1.20 it seems, but not sure what's the solution here

JoshVanL:
@yaron2 I have reverted the `go.mod` to use `1.20` as the minimum version which looks to at least compile and keep compatibility.

JoshVanL:
@yaron2 looking at this more- because dapr/dapr now has a `go.mod` of `1.21` the `go mod tidy` & diff check will always fail. One option is disabling the diff check on older go versions.

.github/workflows/test-on-push.yaml
```diff
@@ -54,4 +54,8 @@ jobs:
           args: --timeout=10m0s
 
       - name: Run go mod tidy check diff
+        if: strategy.job-index == 0
         run: make modtidy check-diff
+      - name: Run go mod tidy
+        if: strategy.job-index != 0
+        run: make modtidy
```
mikeee, line 61: Can/should we compare against the matrix context or the env var set?

.github/workflows/test-on-push.yaml
mikeee, line 1: I'm not sure why we run `tidy` twice, once with a compatibility check. (First instance L40-41)  Probably should be addressed in another PR though.

yaron2: > I'm not sure why we run tidy twice, once with a compatibility check

Agree

yaron2:
> @yaron2 looking at this more- because dapr/dapr now has a `go.mod` of `1.21` the `go mod tidy` & diff check will always fail. One option is disabling the diff check on older go versions.

I'm ok with disabling the diff check on older versions. I'm merging this PR for now, please follow up.

yaron2 (approved):

PR #578, opened 2024-06-19
Bidirectional Subscriptions

JoshVanL:
Adds support for bidirectional subscriptions to PubSubs. Adds two methods for subscribing- one using a callback and one using an imperative approach. Both giving support to different programming styles or use cases.

Adds example with tests.

Requires daprd built from https://github.com/dapr/dapr/pull/7770


mikeee (dismissed):
Lgtm!

mikeee (approved):
lgtm!

mikeee:
> the validation step is failing

It fails as the validation is running the latest version of Dapr rather than the RC.

I've validated with RC2 and it's smooth as butter 🙂 we could set the validation workflow to 1.14.0-rc.2 on the release branch if that helps push this through.

yaron2:
Can't argue with "smooth as butter" :)

marcduiker:
@holopin-bot @JoshVanL Thanks Josh!

PR #584, opened 2024-06-27
Updates dapr/dapr go mod to current master head

JoshVanL:
Updates go to 1.22.3


go.mod
```diff
@@ -1,34 +1,36 @@
 module github.com/dapr/go-sdk
 
-go 1.21.8
+go 1.22.3
+
+toolchain go1.22.4
```
mikeee, line 5:
```suggestion
```
I don't believe a toolchain directive should be specified

cicoyle (changes requested):
We might want to update this again to align with the dapr version, but also there are conflicts.

Issue #634, opened 2024-10-10
Remove obsolete workflow methods


famarting:
equivalent to https://github.com/dapr/dotnet-sdk/issues/1344

the workflow methods defined here https://github.com/dapr/go-sdk/blob/main/client/workflow.go use a deprecated API and we aim to delete them

instead this client should be used https://github.com/dapr/go-sdk/blob/main/workflow/client.go and any docs should be updated to make use of it

famarting:
this issue can be closed since this has been merged https://github.com/dapr/go-sdk/pull/640

this issue does not consist of literally removing the workflow methods, but instead deprecating them and stop advising users to use them

mikeee:
IMO - idiomatic deprecation comments should be added to the relevant client methods before we bottom this issue out completely. e.g.

`client/workflow.go`
```go
...
// StartWorkflowBeta1 starts a workflow using the beta1 spec.
// 
// Deprecated: Please use the workflow client (github.com/dapr/go-sdk/workflow).
// These methods of managing workflows are no longer supported and will be removed in the
// next release.
...
```

Wdyt @famarting?

famarting:
agree agree, will send a PR now adding these comments everywhere

mikeee:
Looks great, thanks for sorting this out 🚀

PR #640, opened 2024-10-18
workflow examples: remove use of deprecated functions

famarting:
# Description

implementation for https://github.com/dapr/go-sdk/issues/634

- updates the workflow examples to only use the recommended client
- adds a Close() function to the workflows client 

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


workflow/client.go
```diff
@@ -211,3 +215,7 @@ func (c *Client) PurgeWorkflow(ctx context.Context, id string) error {
 	}
 	return c.taskHubClient.PurgeOrchestrationState(ctx, api.InstanceID(id))
 }
+
+func (c *Client) Close() {
+	_ = c.conn.Close()
```
mikeee, line 220: I think logging at the debug level here would be helpful rather than discarding any underlying errors from closing the grpc connection

famarting:
its not clear to me if the current test failures are due to the changes in this PR

examples/service/Makefile
```diff
@@ -31,7 +31,7 @@ client: ## Runs the uncompiled example client code
 custom-grpc-client: ## Runs the uncompiled example custom grpc client code
 	dapr run --app-id custom-grpc-client \
 		 -d ./config \
-		 --dapr-http-max-request-size 41 \
+		 --max-body-size 4Mi \
```
mikeee, line 34: Looks like the sidecar was terminating following the updated arg `--max-body-size`. This change should be fine... the validation workflow was pinned to a previous dapr runtime version (1.14.1) but this shouldn't have made a difference, it looks like the CLI hasn't implemented the required changes to accommodate the deprecation of `--dapr-http-max-request-size`

My recommendation is to revert this change until the CLI has implemented this.

examples/service/README.md
```diff
@@ -91,7 +91,7 @@ expected_stdout_lines:
 ```bash
 dapr run --app-id custom-grpc-client \
 		 -d ./config \
-		 --dapr-http-max-request-size 41 \
+		 --max-body-size 4Mi \
```
mikeee, line 94: As above

famarting:
should be all good now

mikeee (approved):

Issue #541, opened 2024-04-03
Workflow - Implement custom retry policies


mikeee:
**Is your feature request related to a problem? Please describe.**

Implement custom retry policies for workflows
https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-features-concepts/#retry-policies

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

famarting:
I'm giving it a try

PR #644, opened 2024-10-25
worflows: activity retry policy

famarting:
# Description

first part of https://github.com/dapr/go-sdk/issues/541

also updated one of the existing examples to show how activity retry policies can be used

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


cicoyle (approved):
lgtm

workflow/activity_context.go
```diff
@@ -61,6 +71,26 @@ func ActivityRawInput(input string) callActivityOption {
 	}
 }
 
+func RetryPolicy(policy ActivityRetryPolicy) callActivityOption {
```
mikeee, line 74: Would this be better suited as `ActivityRetryPolicy`?

workflow/activity_context.go
```diff
@@ -61,6 +71,26 @@ func ActivityRawInput(input string) callActivityOption {
...
+	return func(opts *callActivityOptions) error {
+		opts.retryPolicy = &policy
+		return nil
+	}
+}
+
+func (opts *callActivityOptions) getRetryPolicy() *task.ActivityRetryPolicy {
```
mikeee, line 81: As above

mikeee (commented):
Looks like a solid implementation of retry policies for activities, just one potential signature refactor to accommodate the future addition of workflow retry policies. Also an ask - would it be possible to implement a simple activity that does continually fail - to test the retry logic?

famarting:
good to go

famarting:
can we merge?

mikeee:
Can you please add a test for the activity to retry - I think it's crucial to validate this so we have some visibility if it breaks

mikeee (commented):
Can you please add a test for the activity to retry - I think it's crucial to validate this so we have some visibility if it breaks

examples/workflow/main.go
mikeee, line 1: The new activity hasn't been registered and isn't being invoked

mikeee (changes requested):
Thanks for adding tests, just need to register the new activity

examples/workflow/main.go
famarting: oops 😄

mikeee (approved):
lgtm

PR #648, opened 2024-11-05
Add support for conversation api

yaron2:
Updates to latest Dapr runtime version + support for the new Conversation API


yaron2:
@mikeee was working on a similar implementation at the same time using function options which I prefer, closing this in favor of the upcoming PR

Issue #660, opened 2024-12-09
Topic subscriptions should accept interfaces instead of functions


mfcollins3:
I'm doing a spike for a service that is using the transactional outbox pattern supported by the State Management and Pub/Sub building blocks. I have a gRPC service that exposes a CRUD API and I am implementing the Create operation first. My gRPC specification looks like this:

```protobuf
message CreateActivityRequest {
    message Activity {
        string title = 1;
    }

    Activity activity = 1;
}

message CreateActivityResponse {
    message Activity {
        string id = 1;
        string title = 2;
    }

    Activity activity = 1;
}

service Activities {
    rpc CreateActivity (CreateActivityRequest) returns (CreateActivityResponse) {}
}
```

The gRPC service implementation receives the request, generates a UUID (v7), writes the record to the state store in a transaction, and returns the response. I'm using Redis as the state store during my spike.

```go
type activity {
    ID string `json:"id"`
    Title string `json:"title"`
}

type activitiesServer struct {
    pb.UnimplementedActivitiesServer

    client client.Client // <-- Dapr client
}

func NewServer(client client.Client) pb.ActivitiesServer {
    return &activitiesServer{
        client: client,
    }
}

func (s *activitiesServer) CreateActivity(
    ctx context.Context,
    req *pb.CreateActivityRequest,
) (*pb.CreateActivitiesResponse, error) {
    id, err := uuid.NewV7()
    if err != nil { return nil, err }

    activity := activity{
        ID: id.String(),
        Title: req.GetActivity().GetTitle(),
    }
    jsonActivity, err := json.Marshal(activity)
    if err != nil { return nil, err }

    op := &client.StateOperation{
        Type: client.StateOperationTypeUpsert,
        Item: &client.SetStateItem{
            Key: activity.ID,
            Value: jsonActivity,
        },
    }
    if err := s.client.ExecuteStateTransaction(
        ctx,
        "statestore",
        map[string]string{},
        []*client.StateOperation{op},
    }; err != nil {
        return nil, err
    }

    return &pb.CreateActivityResponse{
        Activity: &pb.CreateActivityResponse_Activity{
            Id: activity.ID,
            Title: activity.Title,
        },
    }, nil
}
```

My service is implementing a subscriber to receive the activity and I want to write it to an Azure SQL database (I'm using SQL Server with Docker Compose in my spike, but it'll eventually be Azure SQL):

```go
func activityHandler(
    ctx context.Context,
    e *common.TopicEvent,
[...]

PR #661, opened 2024-12-09
Introduce interface for topic event subscriber

mfcollins3:
# Description

I introduced the `TopicEventSubscriber` interface and added an `AddTopicEventSubscriber` method to the `Server` interface. `TopicEventSubscriber` will allow more complex subscribers to be registered that are using structs so that dependencies like `sql.DB` objects or other parameters can be provided to subscribers and used to process incoming events.

I modified `TopicRegistrar` to register and store `TopicEventSubscriber` instead of `TopicEventHandler` in the `TopicRegistration struct. I created an adapter for `TopicEventHandler` so that it implements the new `TopicEventSubscriber` interface and can continue to be used without modification.

* I introduced the `TopicEventSubscriber` type instead of rewriting the `TopicEventHandler` type to not break existing code. I used the pattern that `net/http` uses for `Handler` and `HandlerFunc`.
* I did not define any new tests, but I modified the existing tests to work. Given the way that I changed `TopicRegistration` and making `TopicEventHandler` implement `TopicEventSubscriber`, I didn't think that any additional tests were necessary.

## Issue reference

Please reference the issue this PR will close: #660

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation

### DIFF
--- a/daprdocs/content/en/go-sdk-docs/go-service/grpc-service.md
+++ b/daprdocs/content/en/go-sdk-docs/go-service/grpc-service.md
@@ -83,6 +83,35 @@ if err != nil {
 }
 ```
 
+You can also create a custom type that implements the `TopicEventSubscriber` interface to handle your events:
+
+```go
+type EventHandler struct {
+	// any data or references that your event handler needs.
+}
+
+func (h *EventHandler) Handle(ctx context.Context, e *common.TopicEvent) (retry bool, err error) {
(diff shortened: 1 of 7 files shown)


mikeee (approved):
lgtm

yaron2:
Thanks for your contribution!

marcduiker:
@holopin-bot @mfcollins3 Thank you! Here's a digital badge as a small token of appreciation.

PR #685, opened 2025-02-05
add deadlettertopic support to non-streaming subscriptions

yaron2:
Adds missing deadletter topic support for non-streaming subscriptions.


mikeee (approved):
Lgtm

Issue #355, opened 2023-01-29
Transparent not forwarded via PubSub


dkrizic:
I don't know if this is a *bug* or *feature request*. I have two services "server" and "echo". Server sends send a message via pubsub to echo. Both services are fully instrumented with OpenTelemetry. A call to server give me the following trace

![image](https://user-images.githubusercontent.com/1181349/215324386-2af1f5c2-8067-4e43-8da3-a361c6b3f9cd.png)

The last method SendNotifcation forwards the instrumented ctx to the DAPR client:

```
func (n *Sender) SendNotification(ctx context.Context, message []byte) error {
	ctx, span := otel.Tracer("sender").Start(ctx, "SendNotification")
	defer span.End()

	llog := log.WithFields(log.Fields{
		"pubsubName": n.PubSubName,
		"topicName":  n.TopicName,
		"message":    string(message),
	})
	llog.Debug("Sending sender")
	err := n.client.PublishEvent(ctx, n.PubSubName, n.TopicName, message)
	if err != nil {
		llog.WithError(err).Warn("Unable to send sender")
		return err
	}
	return nil
}
```

But the transported message has a new traceid:

```
{
    "data": "{\"Before\":{\"Id\":\"77\",\"Title\":\"Title\",\"Description\":\"The description\",\"Status\":\"\"},\"After\":{\"Id\":\"77\",\"Title\":\"Title\",\"Description\":\"The description\",\"Status\":\"\"},\"ChangeType\":\"CREATE\"}",
    "datacontenttype": "text/plain",
    "id": "b946b14e-1770-4316-8483-9eca55602318",
    "pubsubname": "todo-pubsub",
    "source": "todo",
    "specversion": "1.0",
    "time": "2023-01-29T11:22:03Z",
    "topic": "todo",
    "traceid": "00-f4905fb842564c59ca0905e18f408be7-24f05a94522b81d3-01",
    "traceparent": "00-f4905fb842564c59ca0905e18f408be7-24f05a94522b81d3-01",
    "tracestate": "",
    "type": "com.dapr.event.sent"
}
```

Receiving the trace works again:

```
![image](https://user-images.githubusercontent.com/1181349/215324484-06a6c0f0-a47f-4b46-a311-b7c54fcc95b1.png)
```

But alltogether I have two traces:

![image](https://user-images.githubusercontent.com/1181349/215324508-fd33d191-cf73-468b-94e1-4ebf9f7c918f.png)

Questions:

* Is this feature missing?
[...]

PR #717, opened 2025-04-08
add TraceID and TraceParent handling for grpc and client

majkio:
# Description

Added support for extracting TraceId and TraceParent from Cloud Events Extension

## Issue reference

Please reference the issue this PR will close: [#355](https://github.com/dapr/go-sdk/issues/355)

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/client/subscribe.go
+++ b/client/subscribe.go
@@ -25,6 +25,7 @@ import (
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
 	"github.com/dapr/go-sdk/service/common"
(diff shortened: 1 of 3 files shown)


mikeee (commented):
I think it'd be good to consolidate the getStringValueFromExtension() helper function to an internal package e.g. `internal/structpb` so that we're not duplicating code unnecessarily. I'd also in an ideal world like to see some debugging to differentiate between whether a key has a value declared as "" or whether the key is missing alltogether. What do you think?

tthophan (commented):
**Decision:** `comment` · **Confidence:** 95%

Adds `TraceID` and `TraceParent` extraction from `TopicEvent` extensions. Duplicated helper function `getStringValueFromExtension` in two packages.

client/subscribe.go
```diff
@@ -297,3 +300,18 @@ func (s *Subscription) closeStreamOnly() error {
 	}
```
tthophan, line 300: Move `getStringValueFromExtension` to `service/common` or `internal` to avoid duplication with `service/grpc/topic.go`.

service/grpc/topic.go
```diff
@@ -183,3 +186,18 @@ func getCustomMetadataFromContext(ctx context.Context) map[string]string {
 func (s *Server) OnBulkTopicEventAlpha1(ctx context.Context, in *runtimev1pb.TopicEventBulkRequest) (*runtimev1pb.TopicEventBulkResponse, error) {
 	panic("This API callback is not supported.")
 }
+
```
tthophan, line 189: Duplicated helper. Move to shared location.

PR #722, opened 2025-04-22
fix api token initialization

famarting:
# Description

I noticed that currently, if an API token is provided, it is not used to create the grpc connection, it is only used in subsequent requests after the connection is stablished.

This PR fixes that forcing the API token env var to be read when initializing the authToken holder

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


client/client.go
```diff
@@ -435,14 +430,26 @@ func newClientWithConnection(conn *grpc.ClientConn, authToken *authToken) Client
...
+// it is crucial to correctly initialize the dapr client with the API token from the environment variable
 type authToken struct {
 	mu        sync.RWMutex
 	authToken string
 }
 
+func newAuthToken() *authToken {
+	apiToken := os.Getenv(apiTokenEnvVarName)
+	if apiToken != "" {
+		logger.Println("client uses API token")
```
mikeee, line 446:
```suggestion
		logger.Println("API Token loaded from the environment variable")
```

mikeee (commented):
Minor nit but looks good to me!

mikeee (approved):

PR #728, opened 2025-05-21
Enable gRPC keep-alive by default

lrascao:
# Description

Ensures intermediary proxies/gateways (AWS NLB notably) don't silently close connections after a configurable inactivity timeout expires.

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -35,6 +35,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 
 	pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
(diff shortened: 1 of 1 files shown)


mikeee (commented):
I'm not a fan of adding env vars endlessly to address config options as I'm partial to the functional options pattern but I do think that having both would provide the best flexibility. I propose adding options to NewClient().

I think it would be pertinent to include some documentation of this as well, at the moment I'd hazard a guess that our sparse unexported env var declarations aren't the best for surfacing this config option.

Also raising the question - do you think an incorrect value provided should cause a failure to create a new client? Instead I think defaulting to the keepalive package's 10s when provided an invalid value and logging an error would provide less friction.

client/client.go
```diff
@@ -362,6 +370,13 @@ func NewClientWithAddressContext(ctx context.Context, address string) (client Cl
 		grpc.WithBlock(), //nolint:staticcheck
 		authTokenUnaryInterceptor(at),
 		authTokenStreamInterceptor(at),
+		grpc.WithKeepaliveParams(
+			keepalive.ClientParameters{
```
tthophan, line 374: Incorrect duration conversion. `time.Duration(keepAliveSeconds)` treats the integer as nanoseconds. Use `time.Duration(keepAliveSeconds) * time.Second` instead.

tthophan (changes requested):
**Decision:** `request_changes` · **Confidence:** 95%

The PR enables gRPC keep-alive by default to prevent connection termination by intermediary proxies. The implementation correctly uses `time.Duration` for the keep-alive interval, but the `time.Duration(keepAliveSeconds)` cast is incorrect as `keepAliveSeconds` is an integer representing seconds, not nanoseconds.

PR #737, opened 2025-06-18
Exposes FailurePolicy to the Jobs api

acroca:
# Description

https://github.com/dapr/dapr/pull/8785 has been merged into master. That PR exposes `FailurePolicy` field in the Jobs API

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


cicoyle (approved):
Thanks for the name change to `jobs` for consistency across dapr repos - I've been wanting to update that for a while now 🙌🏻

client/jobs.go
```diff
@@ -0,0 +1,158 @@
...
+	return &JobFailurePolicyDrop{}
+}
+
+type Job struct {
+	Name          string
+	Schedule      string // Optional
+	Repeats       uint32 // Optional
+	DueTime       string // Optional
+	TTL           string // Optional
+	Data          *anypb.Any
```
JoshVanL, line 80: Optional fields should be pointers.

client/jobs.go
```diff
@@ -0,0 +1,158 @@
...
+	if job.DueTime != "" {
+		jobRequest.DueTime = &job.DueTime
+	}
+
+	if job.TTL != "" {
+		jobRequest.Ttl = &job.TTL
+	}
+
+	if job.FailurePolicy != nil {
+		jobRequest.FailurePolicy = job.FailurePolicy.GetPBFailurePolicy()
```
JoshVanL, line 109: Assigned pointers to pointers. We shouldn't empty string match.

client/jobs.go
```diff
@@ -0,0 +1,158 @@
...
+	return err
+}
+
+// GetJobAlpha1 retrieves a scheduled job.
+func (c *GRPCClient) GetJobAlpha1(ctx context.Context, name string) (*Job, error) {
+	// TODO: Name validation
+	resp, err := c.protoClient.GetJobAlpha1(ctx, &runtimepb.GetJobRequest{
+		Name: name,
+	})
+	log.Println(resp)
```
JoshVanL, line 123: remove.

client/jobs.go
```diff
@@ -0,0 +1,158 @@
...
+		},
+	}
+	if f.maxRetries != nil {
+		policy.Policy.(*commonpb.JobFailurePolicy_Constant).Constant.MaxRetries = f.maxRetries
+	}
+	if f.interval != nil {
+		policy.Policy.(*commonpb.JobFailurePolicy_Constant).Constant.Interval = &durationpb.Duration{Seconds: int64(f.interval.Seconds())}
+	}
+	return policy
+}
```
JoshVanL, line 50: Why and what is this?

client/jobs.go
```diff
@@ -0,0 +1,158 @@
...
+		Policy: &commonpb.JobFailurePolicy_Constant{
+			Constant: &commonpb.JobFailurePolicyConstant{},
+		},
+	}
+	if f.maxRetries != nil {
+		policy.Policy.(*commonpb.JobFailurePolicy_Constant).Constant.MaxRetries = f.maxRetries
+	}
+	if f.interval != nil {
+		policy.Policy.(*commonpb.JobFailurePolicy_Constant).Constant.Interval = &durationpb.Duration{Seconds: int64(f.interval.Seconds())}
+	}
```
JoshVanL, line 48: Can this not be built in a better way to avoid type casting the proto pointer?

client/jobs.go
acroca: `JobFailurePolicyConstant` is the exposed type the SDK users can use to assign FailurePolicies. This function then converts this exposed type into the corresponding policy from the `commonpb` package.

client/jobs.go
```diff
@@ -0,0 +1,156 @@
...
+		Data:     job.Data,
+		Schedule: job.Schedule,
+		Repeats:  job.Repeats,
+		DueTime:  job.DueTime,
+		Ttl:      job.TTL,
+	}
+
+	if job.FailurePolicy != nil {
+		jobRequest.FailurePolicy = job.FailurePolicy.GetPBFailurePolicy()
+	}
```
mikeee, line 103: This code doesn't read well, perhaps just a helper function is all that is needed?

```go
func validateFailurePolicy(job.FailurePolicy) (FailurePolicy, err)
```

examples/jobs/main.go
```diff
@@ -49,13 +49,16 @@ func main() {
...
+	interval := time.Second * 3
 	job := daprc.Job{
 		Name:     "prod-db-backup",
 		Schedule: "@every 1s",
 		Repeats:  10,
 		Data: &anypb.Any{
 			Value: jobData,
 		},
+		FailurePolicy: daprc.NewFailurePolicyConstant(&maxRetries, &interval),
 	}
```
mikeee, line 62: I think from this example it's clear that using the addresses of an int or duration adds unnecessary faff in my opinion

client/jobs.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (f *JobFailurePolicyDrop) GetPBFailurePolicy() *commonpb.JobFailurePolicy {
+	return &commonpb.JobFailurePolicy{
+		Policy: &commonpb.JobFailurePolicy_Drop{
+			Drop: &commonpb.JobFailurePolicyDrop{},
+		},
+	}
+}
+
+func NewFailurePolicyConstant(maxRetries *uint32, interval *time.Duration) FailurePolicy {
```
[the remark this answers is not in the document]
acroca: I'll actually remove those 'constructor' functions, as they are now unnecessary (those types were not exposed previously). I think it'll cleaner just to provide the appropriate struct directly, like the following:
```go
&Job{
	Name:     "name",
	Schedule: &schedule,
	Repeats:  &repeats,
	DueTime:  &dueTime,
	TTL:      &ttl,
	Data:     nil,
	FailurePolicy: &JobFailurePolicyConstant{
		maxRetries: &maxRetries,
		interval:   &interval,
	},
}
```
What do you think?

client/jobs.go
acroca: I'm not sure I understand the issue, do you mind adding more detail to your concen?
The `jobRequst.FailurePolicy` is optional, as defined in the proto. This code is just making sure the field remains unset if `job.FailurePolicy` is also not set. and if it's set, it calls its `GetPBFailurePolicy` to get the pb equiavalent.

mikeee: It's more a nit than anything substantial, feel free to dismiss but I just don't think that this code is readable. I'd rather see a helper function that does the same functionality rather than assigning a field and following it up with methods. 

Reading through it again it does make sense what you're doing, I'd potentially like to see a functional options pattern implemented rather than passing empty structs here. e.g.:

```go
func WithJobFailurePolicyConstant(maxRetries int, interval time.Duration) func(*ScheduleJobOption) {
    // Logic
}

func WithJobFailurePolicyDrop() func(*ScheduleJobOption) {
    // Logic
}
```

Example client usage:

```go
err = client.ScheduleJobAlpha1(ctx, &job, daprc.WithJobFailurePolicyConstant(4, time.second * 5))
```

This is equally not a breaking change despite the function signature (ScheduleJobAlpha1) changing.

Wdyt?

client/utils.go
```diff
@@ -30,3 +36,20 @@ func isCloudEvent(event []byte) bool {
...
 }
+
+// toProtoDuration converts a time.Duration to a protobuf duration.
+func toProtoDuration(d time.Duration) *durationpb.Duration {
+	nanos := d.Nanoseconds()
+	secs := nanos / 1e9
+	nanos -= secs * 1e9
+
+	// conversion check - gosec ignored below for conversion
+	if nanos <= int64(math.MinInt32) && nanos >= int64(math.MaxInt32) {
```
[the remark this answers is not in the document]
mikeee: Makes sense... sorry, my bad I missed the refactoring part, thank you for fixing!

client/jobs.go
acroca: I think it's a good suggestion, my only concern with this approach would be the inconsistency between scheduling a job and fetching a job.
If we had those functional options, creating a job with a specific failure policy, the job object itself would not have any failure policy set, as it'll be configured through the functional options. But when fetching a job, the failure policies will come as an attribute in the job object.

I find the current approach more straight forward, treating the failure policy as a normal field in the job object.
Also, if we decide to use functional options for the failure policy, why not doing it for the rest of the fields too? many of those fields are optional, this pattern would be useful for them too.

What do you think?

acroca: Understood, makes sense, I'll do this change and see how it looks. 

Note that the `maxRetries` and `interval` of the constant failure policy are optional, so it'll need to look like the following:
```
daprc.ScheduleJobAlpha1(ctx, 
    daprc.NewJob(
        "jobName",
        daprc.WithJobSchedule("@daily")
        daprc.WithJobConstantFailurePolicy(),
        daprc.WithJobConstantFailurePolicyMaxRetries(4),
        daprc.WithJobConstantFailurePolicyInterval(time.second * 5),
    ),
)
```

mikeee (commented):
Looking good, would be good to see some test assertions for those functional options just to cover any future upstream changes for this Alpha api

acroca:
@mikeee What do you think now?

mikeee (changes requested):
@acroca Looking good, I think the linter isn't happy with the way you're accessing the proto fields but aside from that it looks fine to me.

Could you please address two things also:
- Deps issue whereby the JobFailurePolicy is not found, assumign this is a result of whichever commit hash you've used
- Update the workflow (https://github.com/dapr/go-sdk/blob/main/.github/workflows/validate_examples.yaml) rename the 'dist-scheduler' item to 'jobs'

mikeee (approved):
Lgtm- the tests are flaky when initialising dapr.

acroca:
Cool, what's left to merge this PR?

Issue #749, opened 2025-07-21
Decouple go-sdk and dapr/dapr repos by compiling protos


tscolari:
**Is your feature request related to a problem? Please describe.**

Currently `dapr/go-sdk` depends on `dapr/dapr` just because of the exposed protos.
This creates unnecessary coupling of versions, in particular for projects that use both repositories.

**Describe the solution you'd like**

In my opinion, in an ideal world all generated protos in `dapr/dapr` would be internal/non-exported. Which is the same limitation other SDKs have to deal with, and also what makes them more robust in terms of dependencies. 

There are many ways to accomplish compiling the protos in the go-sdk repo, but I believe probably the best one would be to introduce [buf](https://buf.build/) to the main dapr/dapr repo, which would make it easier to compile the protos here - with the benefit of lock files and so on. Potentially enabling other sdks in the process.

**Describe alternatives you've considered**

There are other ways for compiling the protos in the repo, like fetching them on ci, or using submodules. But probably buf is the more straight forward way.

**Additional context**

PR #751, opened 2025-07-22
Remove dependency to dapr dapr

tscolari:
# Description

This is an attempt to remove the dependency between go-sdk and dapr/dapr.
The main point here is to generate the protos from the remote repository, and refer to them internally only.

There should be no need to pull the whole of dapr/dapr to my go application using the SDK just to use the protos.

There's *a blocker here* though, it seems that dapr/dapr manually add fields to the generated protos, making the generated ones incompatible in a way. The current issue is with the addition of the `CryptoRequest` and `CryptoResponse`.
I suspect some of this is duplicated in other repos, probably the same needs to happen here.

## Issue reference
https://github.com/dapr/go-sdk/issues/749

Please reference the issue this PR will close: #749

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


tscolari:
Ideally codecov should exclude the generated path `./internal/proto` :thinking:

tscolari:
I've created an issue on `dapr/dapr`: https://github.com/dapr/dapr/issues/8901, this would make things simpler here (and potentially for other SDKs too).

mikeee:
I think this should wait until a general consensus is reached with other SDK maintainers before we go full steam ahead with this change.

Makefile
```diff
@@ -61,3 +61,34 @@ check-diff:
...
+		--path dapr/proto/common/v1 \
+		--path dapr/proto/runtime/v1 \
+		"https://github.com/dapr/dapr.git#commit=$$COMMIT"
+
+.PHONY: proto-update
+proto-update:
+	@echo "Updating Dapr to latest commit..."
+	@git ls-remote https://github.com/dapr/dapr.git HEAD | cut -f1 > .dapr-proto-ref
+	@echo "Updated .dapr-proto-ref to: $$(cat .dapr-proto-ref)"
+	@$(MAKE) proto
```
mikeee, line 82: I'd like to see a git tag or hash passed here ideally

PR #771, opened 2025-09-01
Docs: Update markdown based shortcode for card elements

marcduiker:
# Description

This change fixes the rendering in the 1.16 docs. The inner `card` element contains Markdown formatting and should therefore be use the `%`. The card is then rendered as HTML so the outer `cardpanel` should remain using `<>`.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


marcduiker:
Replaced with #772

PR #782, opened 2025-10-02
Jobs: don't force data payload

JoshVanL:
There is no requirement for a job to have a data payload. Remove this check on the jobs client.


acroca (approved):

mikeee (approved):

JoshVanL:
@mikeee please can we merge ?

PR #813, opened 2026-03-02
Add backport workflow

acroca:
Brings https://github.com/dapr/dapr/blob/master/.github/workflows/backport.yaml


.github/workflows/backport.yaml
```diff
@@ -0,0 +1,25 @@
...
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    uses: dapr/.github/.github/workflows/backport.yaml@main
+    secrets:
+      dapr_bot_token: ${{ secrets.DAPR_BOT_TOKEN }}
```
mikeee: Just a thought, do we _need_ to be using the DAPR_BOT_TOKEN for this or can we use the standard actions workflow secret?

mikeee (approved):
I'm fine with the bot token for now but workflows should probably be using the finegrained repo token later

PR #815, opened 2026-03-10
Use shared workflows and actions

acroca:
- Makes use of shared github actions to install the CLI and runtime.
- Adds shared workflows (backport, purge-stale and automerge)


mikeee (changes requested):
This shouldn't be pruned as it's used actively for testing.

acroca:
@mikeee Thoughts? I brought back the inputs to select specific commit SHA for the runtime and CLI. Also kept using `master` when running the `ok-to-test`

tthophan (approved):

